PR #41, opened 2026-03-02
feat: Docker Compose setup with Caddy reverse proxy

gabiudrescu:
Hey there! 👋

First off, great project. I just discovered Claude Code Karma and I love what you've built here. Thanks for putting this together and maintaining it.

I wanted to make it easier to get started without installing Python and Node locally, so I put together a Docker Compose setup. The idea is a single `docker compose up` (or `make up`) and everything runs behind a Caddy reverse proxy on port 8080.

**What's in this PR:**

- Three containers: Caddy (reverse proxy), FastAPI API, SvelteKit frontend
- `~/.claude` is mounted read-only by default so nothing gets modified
- A Makefile for common operations (`make up`, `make dev`, `make health`, etc.)
- A GitHub Actions workflow that builds and pushes images to GHCR on merge to main
- A small change to `frontend/src/lib/config.ts` to handle both browser and SSR API URLs (the existing local dev workflow is unchanged)

I tested this locally and it works well. Dashboard loads, API returns projects, health checks pass. I might be missing edge cases though, so any feedback is welcome.

**Known limitation: live sessions don't show in Docker.**

The live session tracker hook writes state files to `~/.claude_karma/live-sessions/` on the host. In Docker, the API container uses a named volume for `/root/.claude_karma` (because it also writes the SQLite metadata database there), so it doesn't see the host's live session files. Mounting `~/.claude_karma` directly from the host would create SQLite locking conflicts between the container and any native API instance.

A longer-term solution might be replacing SQLite with PostgreSQL or MySQL as a separate container. That would cleanly separate the database from the filesystem, allow the host's `~/.claude_karma/live-sessions/` to be mounted read-only without locking issues, and open the door for multi-instance setups down the road. Just a thought, not sure if that aligns with the project's goals.

No rush on reviewing this. Happy to adjust anything. 🙏

Gabi Udrescu
productowner.ro


gabiudrescu:
Ok. Good to know that. Please reach out if I can help in any way.

gabiudrescu:
Any updates?

PR #50, opened 2026-03-17
fix: support Windows path encoding for projects, sessions, and subagents

linhkk3214:
## Summary

Claude Code on Windows encodes project paths differently from Unix, causing the app to show **no projects, broken sessions, and 500 errors on subagent pages** when running on Windows.

**Root cause:** Claude Code encodes paths by replacing path separators with dashes:
- Unix: `/Users/me/repo` → `-Users-me-repo` (leading `-`)
- Windows: `C:\Code\Tools` → `C--Code-Tools` (drive letter + `--`)

All directory-scan guards used `startswith("-")`, so Windows-encoded dirs like `C--Code-Tools` were silently skipped. Additionally, file reads used the Windows default `cp1252` encoding instead of UTF-8, causing `UnicodeDecodeError` crashes on subagent pages.

### Core path encoding (`api/models/project.py`)
- `encode_path()`: handle Windows absolute paths (`C:\...` → `C--...`)
- `decode_path()`: recover `C:/...` from `C--...` encoded names via regex

### Windows-aware dir filter (`api/utils.py`)
- Add `is_encoded_project_dir(name)` helper — returns `True` for both `-Users-me-repo` (Unix) and `C--Code-Tools` (Windows)
- Replace all `startswith("-")` dir-scan guards across the codebase with the new helper

### Files updated with the new guard:
- `api/db/indexer.py`
- `api/routers/agent_analytics.py`, `history.py`, `projects.py`, `sessions.py`
- `api/services/desktop_sessions.py`, `session_lookup.py`

### UTF-8 encoding fixes
- `api/services/subagent_types.py` — 2 `open()` calls
- `api/models/history.py`, `plugin.py` — `open()` calls

All file reads now use `encoding="utf-8", errors="replace"` to prevent `UnicodeDecodeError` on Windows where Python defaults to `cp1252`.

### Misc
- `api/config.py`: add `localhost:5199` / `127.0.0.1:5199` to CORS allowed origins

## Test plan

- [ ] Run API on Windows, `GET /projects` returns all projects (not empty `[]`)
- [ ] Navigate to a session page — loads without "Session not found" error
- [ ] Navigate to a subagent page — loads without 500 Internal Server Error
- [ ] Existing Unix/Linux/macOS behaviour unchanged (path encoding still produces `-Users-me-repo` format)


JayantDevkar (commented):
## Code Review — PR #50: Windows Path Encoding Fix

Thanks for the contribution! The core Windows path encoding fix is **well-implemented and correct**. The `encode_path`/`decode_path` logic, `is_encoded_project_dir()` guard, and UTF-8 file read fixes all solve the right problems cleanly.

However, there are several issues that need addressing before merge. I've categorized them by severity:

---

### CRITICAL (1) — Must fix

**1. Default API port changed from 8000 → 9005** — `frontend/src/lib/config.ts:21`

The fallback `API_BASE` URL was changed from `http://localhost:8000` to `http://localhost:9005`. This appears to be a local dev config that leaked in. The CLAUDE.md, README, and all project docs reference port 8000. This will break the app for every user who clones and runs without setting `PUBLIC_API_URL`.

---

### HIGH (3) — Should fix

**2. Two test files still use old `startswith("-")` guard**
- `api/tests/api/test_projects.py:92`
- `api/tests/api/test_sessions.py:63`

These test helpers iterate project dirs with the old guard. If someone runs tests against Windows fixtures, these would silently skip Windows-encoded directories, masking bugs.

**3. No unit tests for Windows encode/decode logic**

The PR adds critical new behavior (`encode_path` for Windows, `decode_path` for Windows, `is_encoded_project_dir` for Windows patterns), but there are zero tests for these Windows code paths. Tests should cover:
- `encode_path("C:\\Code\\Tools")` → `"C--Code-Tools"`
- `decode_path("C--Code-Tools")` → `"C:/Code/Tools"`
- `is_encoded_project_dir("C--Code-Tools")` → `True`
- `is_encoded_project_dir("memory")` → `False`

**4. Root-level test scripts and artifacts are repo bloat**

12 files were added to the repo root that don't belong:
- 6 ad-hoc Python test scripts (`quick_browser_test.py`, `test_browser_ui.py`, etc.)
- 3 JSON test result files (`ui_test_results.json`, etc.)
- 2 start scripts (`start.bat`, `start.sh`) using non-standard ports

These aren't integrated into the project's pytest suite, reference hardcoded localhost URLs, and include committed test output data. Will remove before merge.

---

### MEDIUM (4) — Fix when possible

**5. `is_encoded_project_dir` could false-positive on single-letter dirs** — `api/utils.py:37`
The regex `^[A-Za-z]--` would match a dir literally named `a--something`. Unlikely in `~/.claude/projects/` but worth a docstring note.

**6. Redundant `import re` inside functions** — `api/models/project.py:130`, `api/db/indexer.py:586`
`re` is imported at function scope in `decode_path` and `_detect_project_path`. Should be at module top in `project.py`.

**7. `_resolve_manifest_dirs` exported with underscore prefix** — `api/models/plugin.py`
Used across routers despite being "private". Will drop the underscore since it's public API.

**8. Scope creep** — PR bundles ~10 unrelated features (image attachments, timezone fixes, plugin manifests, skills UI, CollapsibleGroup rewrite). Not blocking since most came from the rebase, but worth noting.

---

### LOW (4) — Optional

**9.** `config.ts` JSDoc still says "localhost:8000" (resolves with CRITICAL fix)
**10.** `api/docs/review/sessions/api.md:180` still has old `startswith("-")` pattern
**11.** `start.sh` uses `kill -9` without trying graceful SIGTERM first
**12.** f-string logging in `plugin.py` instead of lazy `%s` formatting

---

I'll address all of these on the branch. The core Windows fix is solid — these are cleanup items to get it merge-ready. 👍

the-non-expert:
Looks good to me now

the-non-expert (approved):

the-non-expert (approved):

PR #57, opened 2026-04-13
fix: resolve subagent type showing _unknown while running

### LINKED ISSUE #56: fix: subagent type shows as _unknown while agent is running
## Bug

When viewing a running subagent in the Agent Session View, its type displays as `_unknown` instead of the correct type (e.g., `oh-my-claudecode:critic`). The type only resolves correctly after the agent finishes.

## Root Cause

`get_all_subagent_types()` in `api/services/subagent_types.py` relies on `tool_result` entries in the parent session JSONL to map `agent_id → subagent_type`. The `tool_result` is only written when an agent **finishes**, so running agents can never be matched via this path.

The agent detail endpoint (`_determine_subagent_type` in `subagent_sessions.py`) calls this function and has no prompt-based fallback — unlike `collect_subagent_info()` which does.

## Fix

Add prompt-based matching as Phase 2.5 in `get_all_subagent_types()`: compare each unmatched agent's initial prompt (first 100 chars, normalized) against Task tool_use input prompts from the parent session. Also filter internal markers (`_unknown`, `_warmup`) from the API response.

## Files Changed

- `api/services/subagent_types.py` — Phase 2.5 prompt matching + 3 helper functions
- `api/routers/subagent_sessions.py` — Convert internal `_` markers to `None` in API response

JayantDevkar:
Closes #56, #58, #59, #60, #61

## What's broken

When you open a subagent's detail page while it's still running, the type badge shows `_unknown` instead of the actual type (e.g. `oh-my-claudecode:critic`). Once the agent finishes, the type magically appears. Beyond that, the agent status system has several gaps compared to session status.

## Issues addressed

| Issue | Gap | Description |
|-------|-----|-------------|
| #56 | Type detection | Agent type shows `_unknown` while running |
| #58 | SSR pre-fetch | Agent page missing server-side live session data |
| #59 | Smart polling | Agent polling doesn't stop after agent completes |
| #60 | Status granularity | No active/idle distinction for running agents |
| #61 | API endpoint | No dedicated agent live-status endpoint |

## Root cause (#56)

We figure out which agent belongs to which Task tool call by matching the `agentId` field inside the `tool_result` message. Problem is, that `tool_result` only gets written to the parent session JSONL **after** the agent finishes. So while it's running, we have no way to link agent → Task → type, and it falls through to the default `_unknown` classification.

### 1. Prompt-based type matching (#56)

Added **Phase 2.5** to `get_all_subagent_types()` — for any agent that's still unmatched after tool_result extraction:
1. Read its first JSONL line to grab the initial prompt (first 100 chars)
2. Normalize it (lowercase, collapse whitespace)
3. Compare against Task tool_use input prompts from the parent session

Also added a guard in `_determine_subagent_type()` to convert internal markers (`_unknown`, `_warmup`) to `null` in the API response.

### 2. SSR live session pre-fetch (#58)

Added `liveSession` fetch to `agents/[agent_id]/+page.server.ts` so the status badge appears immediately on page load instead of waiting for client-side polling.

### 3. Smart polling termination (#59)

Added logic in `ConversationView.svelte` to stop refreshing agent data once the subagent has completed, even if the parent session is still running.

### 4. Active/idle status for agents (#60)

Frontend heuristic: derive `active` vs `idle` from the agent's last message timestamp. No hook changes needed.

### 5. Dedicated agent live-status endpoint (#61)

Added `/agents/{encoded}/{uuid}/agents/{agentId}/live-status` that reads the parent session's state file and returns just the subagent's status entry.

## Testing

- All 1474 existing tests pass
- Verified prompt matching against real session data
- Lint clean (`ruff check`)

## Files changed

| File | What changed |
|------|-------------|
| `api/services/subagent_types.py` | Phase 2.5 prompt matching + 3 helper functions |
| `api/routers/subagent_sessions.py` | Filter `_` markers from API response + live-status endpoint |
| `frontend/src/routes/projects/[project_slug]/[session_slug]/agents/[agent_id]/+page.server.ts` | SSR live session fetch |
| `frontend/src/routes/projects/[project_slug]/[session_slug]/agents/[agent_id]/+page.svelte` | Pass liveSession prop |
| `frontend/src/lib/components/conversation/ConversationView.svelte` | Smart polling termination |
| `frontend/src/lib/components/conversation/ConversationHeader.svelte` | Active/idle heuristic for agents |


api/services/subagent_types.py
```diff
@@ -56,6 +56,21 @@ def get_all_subagent_types(jsonl_path: Path, subagents_dir: Path | None = None)
     for agent_file in agent_files:
         result.update(_extract_types_from_raw_jsonl(agent_file))
 
+    # Phase 2.5: Match unmatched agents by prompt content
+    # Fallback for running agents where no tool_result exists yet.
+    # Compares each unmatched agent's initial prompt against Task input prompts.
+    unmatched_agent_files = [f for f in agent_files if f.stem.removeprefix("agent-") not in result]
```
JayantDevkar, line 62: **Why this exists**: Before this change, an agent only got its type after Phase 1 matched its `agentId` from a `tool_result`. But `tool_result` is only written when the agent finishes. This new phase runs *before* the prefix/content classification (Phase 3-4), so running agents get typed correctly instead of falling through to `_unknown`.

We only build the prompt lookup table when there are actually unmatched agents — no extra JSONL scanning for sessions where everything already resolved.

api/services/subagent_types.py
```diff
@@ -56,6 +56,21 @@ def get_all_subagent_types(jsonl_path: Path, subagents_dir: Path | None = None)
...
+    if unmatched_agent_files:
+        prompt_to_type = _extract_task_prompt_types(jsonl_path)
+        if prompt_to_type:
+            for agent_file in unmatched_agent_files:
+                agent_id = agent_file.stem.removeprefix("agent-")
+                agent_prompt = _get_agent_initial_prompt(agent_file)
+                if agent_prompt:
+                    key = _normalize_key(agent_prompt)
```
JayantDevkar, line 70: **How the matching works**: The Task `input.prompt` field contains the exact text that becomes the agent's first UserMessage. We normalize both sides (lowercase + collapse whitespace) and compare the first 100 chars. This is the same approach `collect_subagent_info()` in `collectors.py:469-471` already uses for the session subagents tab — we're just bringing it to the shared extraction function.

Edge case: if two Task calls have identical first-100-char prompts but different types, the last one wins. In practice this doesn't happen since prompts are unique per spawn.

api/services/subagent_types.py
```diff
@@ -244,3 +259,82 @@ def _classify_by_first_message(agent_file: Path) -> str | None:
...
 
     return None
+
+
+def _normalize_key(text: str) -> str:
+    """Normalize text for matching: lowercase, strip, collapse whitespace."""
+    return " ".join(text.lower().strip().split())
+
+
+def _extract_task_prompt_types(path: Path) -> dict[str, str]:
```
JayantDevkar, line 269: **Three new helpers** — kept minimal and self-contained:
- `_normalize_key()` — same logic as `utils.normalize_key()`, duplicated here to keep `subagent_types.py` dependency-free (it only imports json/logging/re/Path)
- `_extract_task_prompt_types()` — single-pass JSONL scan, same pattern as `_extract_types_from_raw_jsonl()` above
- `_get_agent_initial_prompt()` — reads only the first line of the agent JSONL, same pattern as `_classify_by_first_message()`

api/routers/subagent_sessions.py
```diff
@@ -167,7 +167,11 @@ def _determine_subagent_type(parent_session: Session, agent_id: str) -> Optional
     jsonl_path = parent_session.jsonl_path
     subagents_dir = jsonl_path.parent / jsonl_path.stem / "subagents"
     types = get_all_subagent_types(jsonl_path, subagents_dir)
-    return types.get(agent_id)
+    result = types.get(agent_id)
+    # Convert internal classification markers to None for the API response
+    if result and result.startswith("_"):
```
JayantDevkar, line 174: **Defensive filter**: `get_all_subagent_types()` uses internal markers like `_unknown`, `_warmup`, `_teammate`, `_compact` for classification. These should never leak into the API response. Previously `_unknown` was returned verbatim and the frontend displayed it as a badge label. Now any `_`-prefixed result becomes `null`, which the frontend renders as "Other" via `getSubagentTypeDisplayName()`.

JayantDevkar (commented):
Walkthrough of the changes for review. The core idea: we already have the agent's prompt and the Task input prompt available before the agent finishes — we just weren't matching them.

the-non-expert (approved):

PR #66, opened 2026-04-25
feat: one-command onboarding via npx claude-code-karma

the-non-expert:
## Summary

This PR packages the entire Claude Code Karma stack as a self-contained npm module, enabling one-command installation and launch for any user — no manual Python virtualenv setup, no port configuration, no separate terminal windows.

```bash
npx claude-code-karma
```

That single command will:
1. Find Python 3 on the system
2. Create an isolated virtualenv (`.karma-venv/`) on first run
3. Install all API dependencies (cached on subsequent runs via a stamp file)
4. Detect whether ports 8000 / 5173 are already in use and auto-select free alternatives
5. Start the FastAPI backend and the pre-built SvelteKit frontend in parallel
6. Wait until both services are responding, then open the dashboard in the browser

---

### `bin/start.js` (new)
The CLI entry point. Pure Node.js, no npm dependencies — uses only built-ins (`child_process`, `fs`, `http`, `net`). Key behaviours:

- **Python discovery**: tries `python3` then `python`, exits with a clear message if neither is Python 3
- **Virtualenv management**: creates `.karma-venv/` adjacent to the package root; a `.deps-installed` stamp file prevents pip from re-running on every launch
- **Port conflict resolution**: uses a TCP connect probe (not listen) to detect in-use ports — works correctly on macOS where `listen(0.0.0.0)` can succeed even when `127.0.0.1` is already bound. Falls back to a random free OS-assigned port
- **Process supervision**: prefixes all stdout/stderr from both child processes with `[api]` / `[web]` tags; shuts both down cleanly on `SIGINT` / `SIGTERM` or if either process exits with a non-zero code
- **Startup readiness**: polls both ports over HTTP before printing the dashboard URL, with a 60-second timeout
- **Browser launch**: cross-platform (`open` / `start` / `xdg-open`); can be suppressed with `--no-open`
- **Port overrides**: `--api-port <n>` and `--web-port <n>` flags for users who always want specific ports

### `package.json` (root, new)
Makes the repo publishable as `claude-code-karma` on npm:
- `"bin"` wires `claude-code-karma` → `bin/start.js`
- `"files"` whitelist includes `bin/`, all API source, and `frontend/build/` (the pre-built SvelteKit app)
- `"prepack"` script runs `cd frontend && npm install && npm run build` so the build is always fresh before packing

### `frontend/src/hooks.server.ts` (new)
A SvelteKit server hook that replaces the `%karma_api_base%` placeholder in every SSR HTML response with the actual `KARMA_API_URL` environment variable injected by `bin/start.js` at launch time. This is how the pre-built, static-at-build-time frontend learns which port the API is on at runtime.

### `frontend/src/app.html` (modified)
Adds a single `<script>` tag that seeds `window.__KARMA_API_BASE__` with the placeholder:
```html
<script>window.__KARMA_API_BASE__ = '%karma_api_base%';</script>
```
The placeholder is replaced server-side by `hooks.server.ts` before the page reaches the browser.

### `frontend/src/lib/config.ts` (modified)
Fixes a production bug where all 24 server-side load functions (`+page.server.ts`) would always use the hardcoded `localhost:8000` fallback regardless of what port the API actually started on.

The export now has two branches:
```ts
export const API_BASE: string =
  typeof window === 'undefined'
    // Server-side (SSR): read runtime env var injected by bin/start.js
    ? (process.env.KARMA_API_URL ?? 'http://localhost:8000')
    // Client-side: read window.__KARMA_API_BASE__ injected by hooks.server.ts
    : (window.__KARMA_API_BASE__ && window.__KARMA_API_BASE__ !== '%karma_api_base%'
        ? window.__KARMA_API_BASE__
        : import.meta.env.PUBLIC_API_URL ?? 'http://localhost:8000');
```

The dev-mode behaviour (Vite env var / localhost default) is completely unchanged.

### `.npmignore` (root, new)
Prevents Python build artifacts, the runtime venv, and test files from bloating the npm tarball while explicitly preserving `frontend/build/` (which `.gitignore` normally excludes).

### `frontend/.npmignore` (new)
Overrides `frontend/.gitignore` for npm packaging — keeps `build/`, excludes `src/` and `node_modules/`. Without this file, `npm pack` would use `frontend/.gitignore` and omit the built app entirely.

### `frontend/package.json` + `package-lock.json` (modified)
Adds `@types/node` as a dev dependency. Required for TypeScript to recognise `process.env` in `config.ts` and the pre-existing `hooks.server.ts`. Without it, `npm run check` and `npm run build` both fail with _"Cannot find name 'process'"_.

### `.gitignore` (modified)
- Adds `.karma-venv/` — the runtime virtualenv must not be tracked
- Adds `*.tgz` — prevents accidental commits of `npm pack` output

---

## How the runtime URL injection works end-to-end

```
bin/start.js
  └─ spawns SvelteKit node server with env KARMA_API_URL=http://localhost:<apiPort>

Browser request → SvelteKit node server
  └─ hooks.server.ts reads process.env.KARMA_API_URL
  └─ replaces %karma_api_base% in HTML → window.__KARMA_API_BASE__ = 'http://localhost:<apiPort>'
  └─ +page.server.ts load functions use config.ts → process.env.KARMA_API_URL (SSR path)
  └─ browser JS reads window.__KARMA_API_BASE__ (client path)
```

Both SSR data fetching and client-side API calls always reach the correct port, even when 8000 is occupied.

---

## Test plan

- [ ] `npm run check` passes with 0 errors in `frontend/`
- [ ] With ports 8000 and 5173 free: `npx claude-code-karma` starts both services, opens browser, dashboard loads
- [ ] With port 8000 occupied: startup prints `Port 8000 in use → API on <n>`, dashboard loads and SSR pages return data (not 502)
- [ ] With port 5173 occupied: startup prints `Port 5173 in use → Web on <n>`, dashboard opens on the new port
- [ ] `Ctrl+C` shuts down both services cleanly
- [ ] `npx claude-code-karma --no-open` starts services without opening a browser tab
- [ ] Second run skips pip install (shows `API dependencies... cached`)

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,10 +7,14 @@
 # Dependencies
 node_modules/
 
+# Runtime venv created by bin/start.js
+.karma-venv/
+
 # Build outputs
 dist/
 build/
 out/
+*.tgz
 
 # Vite cache
 .vite/
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,16 @@
+# Python build artifacts (within any included directory)
+**/__pycache__/
+**/*.pyc
+**/*.pyo
+**/*.egg-info/
+
+# Venv created by start.js at runtime
+.karma-venv/
+
+# Frontend test files inside frontend/src/
+frontend/src/tests/
+frontend/src/lib/components/**/__tests__/
+
+# Explicitly include pre-built frontend (overrides build/ in .gitignore)
+!frontend/build
+!frontend/build/**
--- a/bin/start.js
+++ b/bin/start.js
@@ -1,0 +1,237 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const { spawn, execSync } = require('child_process');
+const { existsSync, writeFileSync } = require('fs');
+const path = require('path');
+const http = require('http');
+const net = require('net');
+
+const ROOT = path.join(__dirname, '..');
+const VENV_DIR = path.join(ROOT, '.karma-venv');
+const VENV_BIN = path.join(VENV_DIR, process.platform === 'win32' ? 'Scripts' : 'bin');
+const STAMP = path.join(VENV_DIR, '.deps-installed');
+
+const args = process.argv.slice(2);
+const NO_OPEN = args.includes('--no-open');
+const PREFERRED_API_PORT = parseInt(flagValue('--api-port') || '8000', 10);
+const PREFERRED_WEB_PORT = parseInt(flagValue('--web-port') || '5173', 10);
+
+function flagValue(name) {
+  const i = args.indexOf(name);
+  return i !== -1 ? args[i + 1] : null;
+}
+
+const dim    = (s) => `\x1b[2m${s}\x1b[0m`;
+const green  = (s) => `\x1b[32m${s}\x1b[0m`;
+const cyan   = (s) => `\x1b[36m${s}\x1b[0m`;
+const red    = (s) => `\x1b[31m${s}\x1b[0m`;
+const bold   = (s) => `\x1b[1m${s}\x1b[0m`;
+
+const API_TAG = cyan('[api]');
+const WEB_TAG = green('[web]');
+
+async function main() {
+  console.log('');
+  console.log(bold('  Claude Code Karma'));
+  console.log('');
+
+  // 1. Find Python 3
+  const python = findPython();
+  if (!python) {
+    console.error(red('  ✗ Python 3 not found.'));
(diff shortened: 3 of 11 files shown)


JayantDevkar:
Thanks for this, @the-non-expert! Closing as stale for now — inactive since April and our onboarding flow has shifted since. It's still mergeable, so feel free to reopen/rebase anytime to revive it and we'll take a look. 🙏

PR #75, opened 2026-05-25
feat: live-tracking cleanup + bg-shells/cron data layer + UI

### LINKED ISSUE #76: fix(token-cost): 6 bugs in token/cost calculation and display
## Overview

During an investigation into why the dashboard shows \$100+ costs for sessions on a \$20/month Pro plan, six bugs were found across the token pricing, cost calculation, and analytics aggregation layers. This issue documents all of them with enough context for a future implementer to fix them without re-exploring the codebase.

> **Root cause of user confusion (not a code bug):** The dashboard displays raw Anthropic API pricing — what the same session would cost a pay-as-you-go API customer. Pro subscribers don't pay per-token; Anthropic subsidises usage. \$100 in the dashboard ≠ \$100 charged. A UI disclaimer should be added alongside the cost metric.

---

### File
`api/models/usage.py:189-193`

### Description
The long-context threshold check (200K tokens → 2× input price for Sonnet) uses a `total_input` that **includes `cache_read_input_tokens`**. Claude Code caches conversation context aggressively; a single API call in a heavy session easily reads 200K+ tokens from cache. This flips the entire message to long-context pricing, doubling the cost of uncached input, cache writes, and cache reads alike (all derived from `input_price`).

```python
# current (inflating)
total_input = (
    self.input_tokens + self.cache_creation_input_tokens + self.cache_read_input_tokens
)
is_long = threshold > 0 and total_input > threshold
```

### Blast radius
Every cost calculation for any session that reads >200K tokens from cache per API call. This is the most common heavy-session pattern. Cost can be up to 2× the correct value.

### Proposed fix
Determine whether the threshold in Anthropic's published pricing is based on:
- (a) **total context length** (including cache-read tokens, since they form part of the model's effective context), or
- (b) **uncached prompt tokens only**

If (a): the code is technically correct per billing, but the UI should explain that large cached contexts trigger the surcharge.
If (b): change the threshold check to exclude cache reads:

```python
# option (b) — only count tokens that are "new" to the context
context_size = self.input_tokens + self.cache_creation_input_tokens
is_long = threshold > 0 and context_size > threshold
```

Verify against Anthropic's pricing page (https://www.anthropic.com/pricing) before committing.

---

### File
`api/db/indexer.py:559`

### Description
When persisting subagent rows to `subagent_invocations`, `calculate_cost()` is called without a model argument, silently defaulting to `claude-sonnet-4-6` pricing ($3/M input, $15/M output). If the subagent actually ran on Haiku ($0.80–$1/M input), costs are 3–4× inflated in the subagents/agents analytics tab.

```python
# current — wrong model
usage.calculate_cost()

# correct — pass actual model
usage.calculate_cost(subagent.get_primary_model())
```

### Blast radius
`subagent_invocations.cost_usd` column for every row ever written. Affects:
- Agents tab cost totals
- Plugin cost attribution (plugins.py reads `si.cost_usd` — see Bug 6)
- Any future analytics that joins `subagent_invocations`

Does **not** affect session-level `total_cost` in the `sessions` table (that path is correct).

### Proposed fix
1. Add a helper on the `Agent` model to return the primary model used:
   ```python
   # api/models/agent.py
   def get_primary_model(self) -> Optional[str]:
       models = self.get_models_used()
       return next(iter(models), None)
   ```
2. In the indexer, replace:
   ```python
   usage.calculate_cost()
   ```
   with:
   ```python
   usage.calculate_cost(subagent.get_primary_model())
   ```
3. After deploying, trigger a full re-index so historical rows are rewritten with correct costs.

---

### File
`api/routers/analytics.py:79-80`

### Description
The DB fast path stores `sessions.input_tokens = usage.total_input = input_tokens + cache_creation_input_tokens` (see `indexer.py:438`). A separate column `sessions.cache_creation_tokens` also stores `cache_creation_input_tokens`. The analytics query then sums both and adds them together:

```python
# current — double-counts cache_creation
total_cacheable = (
    totals["total_input"] + totals["total_cache_creation"] + totals["total_cache_read"]
)
# = (input + cache_creation) + cache_creation + cache_read
# = input + 2×cache_creation + cache_read   ← WRONG
```

This inflates the denominator, making the displayed cache hit rate **lower** than the real value.

### Blast radius
Cache hit rate metric on the global analytics page and project analytics page (DB fast path only). The full-parse fallback path (`analytics.py:532-533`) is correct.

### Proposed fix
```python
# correct — total_input already contains cache_creation
total_cacheable = totals["total_input"] + totals["total_cache_read"]
```

One-line fix. No schema change needed.

---

### File
`api/routers/analytics.py:546-554`

### Description
When a session used multiple models, the project analytics path splits the session's aggregate tokens evenly using integer division, then prices each split at that model's rate:

```python
model_count = len(session_models)
for model in session_models:
    split_usage = TokenUsage(
        input_tokens=usage.input_tokens // max(1, model_count),   # even split
        output_tokens=usage.output_tokens // max(1, model_count),
        ...
    )
    total_cost += split_usage.calculate_cost(model)
```

This is doubly wrong:
1. `//` integer division silently drops up to `model_count − 1` tokens per field per session.
2. Even splitting is never accurate — a session typically uses Sonnet for 95% of turns and Haiku for agents, producing severe cost misattribution.

### Blast radius
Project-level `estimated_cost_usd` in the analytics tab, for any project that used more than one model in a session. Session-level `total_cost` (calculated per-message with the real model) is unaffected.

### Proposed fix
The session model already computes per-message cost correctly (`session.py:459`). For analytics, retrieve `session.total_cost` directly instead of re-computing:

```python
# replace the model-split loop with:
total_cost += session.get_total_cost()
```

This avoids the even-split approximation entirely since `get_total_cost()` iterates each `AssistantMessage` with its actual model.

---

### File
`api/routers/plugins.py:606-608`

### Description
Cost attribution for plugins hardcodes Sonnet input/output prices and ignores cache tokens:

```python
# current — hardcoded, no cache
input_cost = (usage.input_tokens / 1_000_000) * 3.0
output_cost = (usage.output_tokens / 1_000_000) * 15.0
session_cost = input_cost + output_cost
```

This ignores:
- Model actually used (Haiku sessions show 3× inflated plugin cost)
- Cache write tokens (priced at 1.25× input)
- Cache read tokens (priced at 0.10× input)
- Long-context pricing

### Blast radius
Plugin tab cost totals only. Does not affect session or project cost.

### Proposed fix
```python
from models.usage import TokenUsage

usage_obj = TokenUsage(
    input_tokens=usage.input_tokens,
    output_tokens=usage.output_tokens,
    cache_creation_input_tokens=usage.cache_creation_input_tokens,
    cache_read_input_tokens=usage.cache_read_input_tokens,
)
session_cost = usage_obj.calculate_cost(session.get_primary_model())
```

Note: `session.get_primary_model()` or `session.get_models_used()` will need to be checked — the variable naming in this router should be verified before patching.

---

### File
`api/routers/analytics.py:92` and `api/db/queries.py:439-440`

### Description
The "Total Tokens" stat on analytics pages is computed as:

```python
total_tokens = totals["total_input"] + totals["total_output"]
# = (input + cache_creation) + output
# cache_read_tokens are NOT included
```

The DB column `sessions.input_tokens` stores `input + cache_creation` but not cache reads. So "Total Tokens" silently omits cache-read tokens from the displayed count, making it look like much less work happened than actually did.

### Blast radius
Display only — cosmetic undercount of total tokens on analytics pages. Does not affect cost calculation.

### Proposed fix
Option A: Update the SQL aggregation to include `cache_read_tokens`:
```sql
-- in db/queries.py query_analytics()
COALESCE(SUM(input_tokens), 0) + COALESCE(SUM(cache_read_tokens), 0) as total_input
```

Option B: Add a separate `total_cache_read_tokens` field to the response and show it as a breakdown in the UI.

Option B is more informative and doesn't break the "total input" semantic.

---

## Non-bug: Pro plan disclaimer

No code change needed for cost accuracy, but add a tooltip or footnote next to every cost metric:

> "Estimated API cost at Anthropic's pay-as-you-go rates. Pro and Max subscribers pay a flat monthly fee and are not charged per token."

This is the single highest-impact UX fix for the user confusion described in this issue.

---

## Files touched by all fixes

| File | Change |
|------|--------|
| `api/models/usage.py` | Bug 1 — threshold check |
| `api/models/agent.py` | Bug 2 — `get_primary_model()` helper |
| `api/db/indexer.py` | Bug 2 — pass model to `calculate_cost()` |
| `api/routers/analytics.py` | Bugs 3 & 4 — cache denominator + model split |
| `api/routers/plugins.py` | Bug 5 — replace hardcoded pricing |
| `api/db/queries.py` | Bug 6 — include cache_read in total_tokens |
| `frontend/src/lib/components/` | UI disclaimer for cost metrics |

## Suggested fix order

1. **Bug 3** (one-liner, no risk)
2. **Disclaimer UI** (highest user-facing impact, no logic change)
3. **Bug 2** (subagent model + re-index)
4. **Bug 4** (replace even-split with `get_total_cost()`)
5. **Bug 5** (plugin pricing)
6. **Bug 1** (verify Anthropic billing semantics first, then patch or document)
7. **Bug 6** (display-only, low priority)

JayantDevkar:
## Summary

Four scoped changes shipped together on one branch:

### 1. Live-tracking + session-UX cleanup (4 commits, original PR scope)
- Full session ID + resume command UX
- Unified live-session writers, dropped dead slug code, fixed subagent timing
- Widened `SubagentState.started_at` to nullable + exposed `duration_ms`
- Dropped the Copy Path button from session header

### 2. Schema v13 — background shells + cron data layer (7 commits)

Models Claude Code's background-shell and cron tool calls so the dashboard can visualize them. Strictly data layer.

**New tables (4):**
- `background_shells` — one row per `Bash{run_in_background:true}` or `Monitor`
- `shell_polls` — one row per `BashOutput`, 4KB excerpt inline
- `cron_jobs` — one row per `CronCreate`, folds in `CronDelete`, 7-day TTL
- `cron_state_snapshots` — optional ground-truth from a PostToolUse hook

**Architectural decisions (after independent DB-expert review):**
- UPSERT on `UNIQUE(tool_use_id)` everywhere — no `INSERT OR REPLACE` cascade churn
- Coherence CHECKs on `(terminated_at, terminated_by)` and `(deleted_at, deleted_via)`
- `cron_fires` NOT persisted — derived on read via croniter, tune the matching window without re-indexing
- Cross-branch safety: new SQL block applied unconditionally in `ensure_schema()`

**Optional hook** (`hooks/cron_state_capture.py`): off by default, PostToolUse that captures `CronCreate`/`CronDelete`/`CronList` results.

**Dependency:** added `croniter>=2.0.0`.

### 3. UI surface (2 commits)

**API routers** — `/shells`, `/shells/project-rollup`, `/cron`, `/cron/project-rollup`, `/sessions/{uuid}/shells`, `/sessions/{uuid}/cron` (all read-only).

**Global pages:**
- `/shells` — ledger/Gantt-swimlane layout with 8-char shell_id labels, status glyphs, color-coded timeline bars, expand-to-detail.
- `/cron` — two-pane layout with honest "reconstructed" framing strip, big mono cron_id headlines, TTL countdown bar, fires timeline.

**Per-session tabs:** new `Shells` and `Cron` tabs on session detail pages (main-session-only). Compact inline lists with expand-to-detail.

**Sidebar nav:** `Shells` (Activity / green) and `Cron` (AlarmClock / yellow) added between Tools and Hooks.

### 4. Token/cost calculation fixes — closes #76 (2 commits)

Six bugs across the pricing, cost calculation, and analytics aggregation layers:

- **Bug 1** (`usage.py`) — Confirmed/documented that the long-context 2× threshold correctly counts *total* context (uncached + cache-creation + **cache-read**) tokens, matching Anthropic billing (the model attends to cache-read tokens too). Comment-only clarification, no behavior change — the threshold intentionally includes cache reads (locked in by `test_long_context_includes_cache_in_threshold`). _(Corrected: an earlier draft of this line claimed the threshold excludes cache reads — that was the inverse of the actual, correct implementation.)_
- **Bug 2** (`indexer.py` + `agent.py`) — Subagent cost stored in DB used default Sonnet pricing regardless of actual model; added `get_primary_model()` and passed it to `calculate_cost()`
- **Bug 3** (`analytics.py`) — Cache hit rate denominator double-counted cache-creation tokens (already folded into `sessions.input_tokens`); one-line fix
- **Bug 4** (`analytics.py`) — Project-level cost split tokens evenly across models via integer division; replaced with `session.get_total_cost()` which prices per-message with the actual model
- **Bug 5** (`plugins.py`) — Plugin router hardcoded Sonnet rates and ignored cache tokens; replaced with `TokenUsage.calculate_cost(model)`
- **Bug 6** (`api-types.ts` + UI) — "Total Tokens" stat excluded cache-read tokens; it now **includes** them. _(Clarification: cache reads are folded into the Total Tokens figure, not yet broken out as a separate stat — a dedicated cache-read display is tracked as a follow-up.)_

## Test plan

- [x] All 129 existing schema/db tests still pass
- [x] 45 new tests in `test_indexer_shells_cron.py` (extraction, idempotency, CHECK constraints, CASCADE, hook ingestion, fire inference, query helpers)
- [x] 28 new router tests (filters, validation, 404s, response shapes)
- [x] Full test suite: 1716 passing, 6 skipped, 0 failing
- [x] Frontend `npm run check`: 0 errors (15 warnings are pre-existing patterns)
- [x] Migration verified: v12 → v13 adds column + flags 2,190 sessions for backfill
- [x] Real-data backfill: 516 background shells + 21 cron jobs extracted from local sessions
- [x] End-to-end browser verification with Playwright:
  - `/shells` renders 200 closed bg shells with ladder + filters working
  - `/cron` renders 21 jobs with **45–49 fires inferred per job** by croniter
  - Per-session Shells tab shows 59 shells for the heaviest session
  - Per-session Cron tab shows scheduled jobs with TTL countdown + honesty banner

### DIFF
--- a/api/db/indexer.py
+++ b/api/db/indexer.py
@@ -162,6 +162,17 @@ def sync_all_projects(conn: sqlite3.Connection) -> dict:
         # Clean up stale sessions (files deleted from disk)
         _cleanup_stale_sessions(conn, projects_dir)
 
+        # Ingest optional cron live-state hook output. Idempotent via
+        # UNIQUE(session_uuid, trigger_event, captured_at).
+        try:
+            from db.indexer_shells_cron import sync_cron_state_snapshots
+
+            inserted = sync_cron_state_snapshots(conn, settings.karma_base)
+            if inserted:
+                logger.info("Ingested %d new cron-state snapshots", inserted)
+        except Exception as e:
+            logger.warning("cron-state snapshot ingestion failed: %s", e)
+
         # Update project summary table
         _update_project_summaries(conn)
 
@@ -224,12 +235,23 @@ def sync_project(
     source_encoded_name = project_dir.name if encoded_name_override else None
     stats = {"total": 0, "indexed": 0, "skipped": 0, "errors": 0}
 
-    # Load current mtimes from DB for this project
-    rows = conn.execute(
-        "SELECT uuid, jsonl_mtime FROM sessions WHERE project_encoded_name = ?",
-        (encoded_name,),
-    ).fetchall()
-    db_mtimes = {row["uuid"]: row["jsonl_mtime"] for row in rows}
+    # Load current mtimes + v13 reindex flag from DB for this project.
+    # The flag column may not yet exist on older schemas — fall back to {}.
+    try:
+        rows = conn.execute(
+            "SELECT uuid, jsonl_mtime, needs_shell_cron_reindex "
+            "FROM sessions WHERE project_encoded_name = ?",
+            (encoded_name,),
+        ).fetchall()
+        db_mtimes = {row["uuid"]: row["jsonl_mtime"] for row in rows}
+        needs_backfill = {row["uuid"]: bool(row["needs_shell_cron_reindex"]) for row in rows}
+    except sqlite3.OperationalError:
+        rows = conn.execute(
+            "SELECT uuid, jsonl_mtime FROM sessions WHERE project_encoded_name = ?",
+            (encoded_name,),
+        ).fetchall()
+        db_mtimes = {row["uuid"]: row["jsonl_mtime"] for row in rows}
+        needs_backfill = {}
 
     for jsonl_path in project_dir.glob("*.jsonl"):
         # Skip agent files and non-session files
@@ -244,9 +266,19 @@ def sync_project(
             current_mtime = file_stat.st_mtime
             current_size = file_stat.st_size
 
-            # Skip if mtime hasn't changed
+            # mtime unchanged: skip the full re-index. But if the v13
+            # bg-shells/cron backfill is still pending for this row, run
+            # the lightweight extraction-only pass.
             if uuid in db_mtimes and abs(db_mtimes[uuid] - current_mtime) < 0.001:
-                stats["skipped"] += 1
+                if needs_backfill.get(uuid):
+                    try:
+                        _backfill_shells_cron(conn, jsonl_path, uuid)
+                        stats["indexed"] += 1
+                    except Exception as e:
+                        logger.debug("backfill shells/cron error for %s: %s", uuid, e)
+                        stats["errors"] += 1
+                else:
+                    stats["skipped"] += 1
                 continue
 
             # Index this session
@@ -562,6 +594,55 @@ def _index_session(
                     )
         except Exception as e:
             logger.warning("Error indexing subagent invocations for %s: %s", uuid, e)
+
+    # v13: bg-shells/cron extraction pass. Runs whenever mtime changed
+    # (i.e. whenever _index_session is called) and on backfill via
+    # _backfill_shells_cron(). Wrapped in try/except so a parser bug
+    # cannot break the main indexer.
+    try:
+        from db.indexer_shells_cron import (
+            extract_shells_and_cron,
+            persist_shells_and_cron,
+        )
+
+        shells, polls, crons = extract_shells_and_cron(
+            jsonl_path,
+            uuid,
+            session_ended=session.end_time is not None,
+        )
+        persist_shells_and_cron(conn, uuid, shells, polls, crons)
+    except Exception as e:
+        logger.warning("Error indexing bg-shells/cron for %s: %s", uuid, e)
+
+
+def _backfill_shells_cron(
+    conn: sqlite3.Connection,
+    jsonl_path: Path,
+    session_uuid: str,
+) -> None:
+    """
+    Lightweight pass: extract + persist only bg-shells/cron for an existing
+    session whose mtime hasn't changed but whose needs_shell_cron_reindex
+    flag is set (e.g. post-v13 migration).
+
+    Skips all the heavy Session model parsing in _index_session().
+    """
+    from db.indexer_shells_cron import (
+        extract_shells_and_cron,
+        persist_shells_and_cron,
+    )
+
+    # session_ended: rely on stored end_time rather than parsing the JSONL.
+    row = conn.execute(
+        "SELECT end_time FROM sessions WHERE uuid = ?",
+        (session_uuid,),
+    ).fetchone()
+    session_ended = bool(row and row[0])
+
+    shells, polls, crons = extract_shells_and_cron(
+        jsonl_path, session_uuid, session_ended=session_ended
+    )
+    persist_shells_and_cron(conn, session_uuid, shells, polls, crons)
 
 
 def _detect_project_path(session, encoded_name: str) -> Optional[str]:
--- a/api/db/indexer.py
+++ b/api/db/indexer.py
@@ -28,6 +28,38 @@ sys.path.insert(0, str(Path(__file__).parent.parent))
 
 logger = logging.getLogger(__name__)
 
+# Sessions whose JSONL was modified within this window are treated as
+# potentially still-active, regardless of whether end_time is set.
+# (session.end_time is always equal to the last message timestamp — it is
+# never None for a session with messages, so it cannot be used to distinguish
+# a truly-ended session from an active one.)
+_LIVE_FALLBACK_SECS = 30 * 60  # 30 minutes
+
+
+def _is_session_ended(session_uuid: str, jsonl_mtime: float) -> bool:
+    """
+    Return True when the session has truly ended.
+
+    Primary: read the live-session state file written by the hooks.
+      ENDED   → True   (hooks confirmed shutdown)
+      LIVE / WAITING / STALE → False (session still open)
+      absent  → fall through to heuristic
+
+    Fallback (no live-session file): assume ended if the JSONL has not been
+    touched in the last _LIVE_FALLBACK_SECS seconds.
+    """
+    state_file = Path.home() / ".claude_karma" / "live-sessions" / f"{session_uuid}.json"
+    if state_file.exists():
+        try:
+            import json as _json
+
+            data = _json.loads(state_file.read_text())
+            return data.get("state", "").upper() == "ENDED"
+        except Exception:
+            pass
+    return (time.time() - jsonl_mtime) > _LIVE_FALLBACK_SECS
+
+
 # Module-level state
 _ready = threading.Event()
 _indexing_lock = threading.Lock()
@@ -608,7 +640,7 @@ def _index_session(
         shells, polls, crons = extract_shells_and_cron(
             jsonl_path,
             uuid,
-            session_ended=session.end_time is not None,
+            session_ended=_is_session_ended(uuid, mtime),
         )
         persist_shells_and_cron(conn, uuid, shells, polls, crons)
     except Exception as e:
@@ -632,15 +664,10 @@ def _backfill_shells_cron(
         persist_shells_and_cron,
     )
 
-    # session_ended: rely on stored end_time rather than parsing the JSONL.
-    row = conn.execute(
-        "SELECT end_time FROM sessions WHERE uuid = ?",
-        (session_uuid,),
-    ).fetchone()
-    session_ended = bool(row and row[0])
-
     shells, polls, crons = extract_shells_and_cron(
-        jsonl_path, session_uuid, session_ended=session_ended
+        jsonl_path,
+        session_uuid,
+        session_ended=_is_session_ended(session_uuid, jsonl_path.stat().st_mtime),
     )
     persist_shells_and_cron(conn, session_uuid, shells, polls, crons)
 
--- a/api/db/indexer.py
+++ b/api/db/indexer.py
@@ -556,7 +556,7 @@ def _index_session(
                         subagent_type,
                         usage.total_input,
                         usage.output_tokens,
-                        usage.calculate_cost(),
+                        usage.calculate_cost(subagent.get_primary_model()),
                         duration,
                         subagent.start_time.isoformat() if subagent.start_time else None,
                     ),
--- a/api/db/indexer_shells_cron.py
+++ b/api/db/indexer_shells_cron.py
@@ -47,6 +47,26 @@ _SHELL_ID_RE = re.compile(
     re.IGNORECASE,
 )
 
+# Modern Claude Code (post-BashOutput) reports the output file path in the
+# Bash{run_in_background} tool_result and instructs the agent to Read it.
+# Pattern: "Output is being written to: /private/tmp/.../tasks/abc.output."
+_OUTPUT_FILE_RE = re.compile(
+    r"Output is being written to:\s*(\S+\.output)",
+    re.IGNORECASE,
+)
+
+# When a background shell exits naturally, Claude Code writes a user message
+# whose content is a raw XML string (not a list of blocks):
+#   <task-notification>
+#     <task-id>bx93dg00s</task-id>
+#     <tool-use-id>toolu_01Wqvk3U47...</tool-use-id>
+#     <status>completed</status>
+#     <summary>... completed (exit code 0)</summary>
+#   </task-notification>
+_TASK_NOTIF_TOOL_USE_ID_RE = re.compile(r"<tool-use-id>\s*(.*?)\s*</tool-use-id>")
+_TASK_NOTIF_STATUS_RE = re.compile(r"<status>\s*(.*?)\s*</status>")
+_TASK_NOTIF_EXIT_CODE_RE = re.compile(r"exit code\s+(\d+)", re.IGNORECASE)
+
 # Cron job IDs are documented as 8-char alphanumeric. Real formats observed:
 #   "task created with ID: a4f9b2c1" or "id": "a4f9b2c1" in JSON.
 _CRON_ID_RE = re.compile(
@@ -107,6 +127,33 @@ def _extract_shell_id(result_text: str) -> Optional[str]:
     return m.group(1) if m else None
 
 
+def _extract_output_file(result_text: str) -> Optional[str]:
+    m = _OUTPUT_FILE_RE.search(result_text)
+    return m.group(1) if m else None
+
+
+def _parse_task_notification(content: str) -> Optional[Dict[str, Any]]:
+    """
+    Parse a <task-notification> string written by Claude Code when a background
+    shell exits naturally. Returns a dict with tool_use_id, terminated_by, and
+    exit_code, or None if the string doesn't match.
+    """
+    if "<task-notification>" not in content:
+        return None
+    tu_m = _TASK_NOTIF_TOOL_USE_ID_RE.search(content)
+    st_m = _TASK_NOTIF_STATUS_RE.search(content)
+    if not tu_m or not st_m:
+        return None
+    status = st_m.group(1).lower()
+    terminated_by = "timeout" if status == "timeout" else "natural"
+    ec_m = _TASK_NOTIF_EXIT_CODE_RE.search(content)
+    return {
+        "tool_use_id": tu_m.group(1),
+        "terminated_by": terminated_by,
+        "exit_code": int(ec_m.group(1)) if ec_m else None,
+    }
+
+
 def _extract_cron_id(result_text: str) -> Optional[str]:
     m = _CRON_ID_RE.search(result_text)
     return m.group(1) if m else None
@@ -155,6 +202,11 @@ def extract_shells_and_cron(
     # Buffered BashOutput rows pending their tool_result (which carries bytes)
     pending_polls: Dict[str, Dict[str, Any]] = {}  # by BashOutput tool_use_id
     polls: List[Dict[str, Any]] = []
+    # Modern Claude Code polls via Read on the output file instead of BashOutput.
+    # Map: output_file_path → parent shell tool_use_id
+    shell_output_files: Dict[str, str] = {}
+    # Buffered Read-poll rows pending their tool_result
+    pending_read_polls: Dict[str, Dict[str, Any]] = {}  # by Read tool_use_id
 
     # Cron, keyed by CronCreate tool_use_id
     crons_by_tool_use: Dict[str, Dict[str, Any]] = {}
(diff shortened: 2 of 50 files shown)


the-non-expert:
## Fix applied: `stop_hook_active` default bug (commit c034434)

**File:** `hooks/_captain_hook_lite.py` — `StopHook.stop_hook_active` default changed `False` → `True`

### What was wrong

The `StopHook` Pydantic model defaulted `stop_hook_active` to `False`. When Claude Code fires a `Stop` hook without this field (e.g. older builds), the tracker evaluated:

```python
if not _get(hook, "stop_hook_active", True):  # True = fallback for raw dict path
    write_state(session_id, "STOPPED", data)
```

With pydantic available (the normal case), `hook.stop_hook_active` returns the model default `False`, so `not False = True` → session incorrectly marked **STOPPED**.

The raw-dict fallback path (`_get(..., default=True)`) had the opposite behavior — field absent → `True` → `not True = False` → session NOT marked STOPPED.

Two code paths, opposite results for the same missing-field case.

### Reproduced locally

```bash
# SessionStart to create the file
echo '{"hook_event_name":"SessionStart","session_id":"test-stop-bug","cwd":"/tmp",...}' \
  | python3 hooks/live_session_tracker.py
# → state: STARTING ✓

# Stop WITHOUT stop_hook_active field (before fix)
echo '{"hook_event_name":"Stop","session_id":"test-stop-bug","cwd":"/tmp",...}' \
  | python3 hooks/live_session_tracker.py
# → state: STOPPED ❌  (should be STARTING)
```

After fix: same payload → state stays `STARTING` ✅

### The fix

`True` is the conservative default: assume a forced-continue is in progress unless Claude explicitly sends `stop_hook_active: false`. One character change.

the-non-expert:
## Review + Fixes by @ayushj

Full review pass on this branch, followed by local testing and fixes. Here's everything that was done.

---

### 🔍 Code Review Findings (inline analysis)

Reviewed the entire diff across `hooks/`, `api/`, and `frontend/`. Key findings:

- **`StopHook.stop_hook_active` default was `False`** — when Claude Code omits the field, the hook should assume it *is* continuing (conservative default). `False` caused sessions to be incorrectly marked STOPPED. Fixed.
- **`live_session_store.py` shallow merge footgun** — `existing.update(updates)` silently drops fields not in `updates`. Worth a second look if partial-update callers are added later.
- **Error count formula in analytics** — minor off-by-one risk worth auditing.
- **CI: `croniter` missing from dev deps** — `infer_cron_fires` degrades silently without it, causing fire-inference tests to assert `0 == 2`. Fixed.
- **CI: ruff import ordering + formatting** — 6 files needed `ruff format`, 3 needed `--select I` import sort fix. Fixed.

---

### 🐛 Bugs Fixed & Pushed

#### `hooks/_captain_hook_lite.py` — StopHook default ([`c034434`](https://github.com/JayantDevkar/claude-code-karma/commit/c034434))
`StopHook.stop_hook_active` defaulted to `False`. When Claude Code omits the field (common), the tracker incorrectly marked the session STOPPED. Changed default to `True` — assume the hook *is* still active unless Claude explicitly says otherwise.

#### `api/pyproject.toml` + ruff formatting ([`ebf70cd`](https://github.com/JayantDevkar/claude-code-karma/commit/ebf70cd), [`a20fc21`](https://github.com/JayantDevkar/claude-code-karma/commit/a20fc21))
Two CI failures:
1. Import ordering (`I001`) in 3 new test files → fixed with `ruff check --select I --fix`
2. `ruff format` diff on 6 files → reformatted
3. `croniter` not in `[dev]` extras → `pip install -e ".[dev]"` in CI skipped it, causing `infer_cron_fires` to return `[]` and 2 fire-inference tests to fail with `assert 0 == 2`

#### `api/db/indexer.py` — live shell detection ([`380722e`](https://github.com/JayantDevkar/claude-code-karma/commit/380722e))
**Root cause:** `session_ended` was computed as `session.end_time is not None`. But `session.end_time` is always set — it equals the last message's timestamp, not a true shutdown event. Every bg-shell was therefore stamped `terminated_by: session_end` the moment its session was first indexed. Nothing could ever appear as RUNNING.

**Fix:** replaced the broken check with `_is_session_ended(uuid, mtime)`:
1. **Primary** — reads `~/.claude_karma/live-sessions/{uuid}.json`. Returns `True` only when `state == "ENDED"` (written by hooks on `SessionEnd`). `LIVE` / `WAITING` / `STALE` → `False`.
2. **Fallback** (no live-session file) — JSONL mtime older than 30 min → ended, otherwise → still live.

Both call sites fixed: `_index_session` and `_backfill_shells_cron`.

---

### 🎨 `/shells` UI — Design Handoff + Bug Fixes ([`09c9e2e`](https://github.com/JayantDevkar/claude-code-karma/commit/09c9e2e))

Applied the design handoff to `/shells` and fixed every issue found during testing:

| Bug | Root cause | Fix |
|-----|-----------|-----|
| **Blank list** — nothing rendered at all | `tool_use_id.slice(0, 8)` → `toulu_01` for every Anthropic tool ID (shared prefix). Svelte's keyed `{#each}` threw `each_key_duplicate` and silently rendered nothing | Use full `tool_use_id` as fallback key |
| **Filter counts frozen** — seg buttons always showed 123/0/123 regardless of active filters | `counts` derived from all `shells`, not the filtered context | Added `crossFiltered` (all filters except status) → `counts` now reflects tool/project/search context |
| **Section bar missing `--project`** | Template only interpolated `--status` and `--tool` | Added `--project=<name>` conditional |
| **Session name not linked** | Plain text `{s.session}` | Now an `<a href="/projects/{encoded_name}/{slug}">` when URL is available |
| **Refresh button was a no-op** | No `onclick` handler | Wired to `invalidateAll()` — actually re-fetches from API |
| **Export button was a no-op placeholder** | No handler, no implementation | Removed |

---

### 🧹 Live Sessions DB Cleanup
Purged 26 stale `ENDED`/`STALE` live-session files from `~/.claude_karma/live-sessions/`, keeping only the active session.

---

### 🐚 Background Shells — Full Feature Pass ([`a15f702`](https://github.com/JayantDevkar/claude-code-karma/commit/a15f702))

Four bugs/gaps found and fixed during end-to-end testing of the `/shells` page:

#### 1. Poll data missing from global `/shells` endpoint
`get_shells_global()` did a bare `SELECT bs.*` — no second pass for `shell_polls`. The per-session endpoint already had `include_polls` logic; the global endpoint was missing it entirely. Added the same `IN (...)` bulk fetch to `get_shells_global()`.

#### 2. Read-based polling (modern Claude Code behaviour)
The indexer looked for `BashOutput` tool calls to create `shell_polls` rows. But current Claude Code no longer uses `BashOutput` — it polls bg shells by calling `Read` on the output file (`/private/tmp/.../tasks/{task-id}.output`). The `BashOutput` path still works for older sessions; added parallel detection for `Read`-on-output-file calls so polls populate for all current sessions.

#### 3. Natural exit detection via `<task-notification>`
When a bg shell exits on its own, Claude Code writes a `user` message to the JSONL whose `content` is a raw XML string:
```xml
<task-notification>
  <tool-use-id>toolu_01...</tool-use-id>
  <status>completed</status>
  <summary>... completed (exit code 0)</summary>
</task-notification>
```
The indexer now parses this: sets `terminated_at` to the exact message timestamp, `terminated_by='natural'`, and `exit_code` from the summary. Shells that finish mid-session are now correctly shown as CLOSED with accurate timestamps rather than staying RUNNING until the entire session ends.

#### 4. Kill Shell button — fully implemented
- Schema v18: `output_file_path TEXT` column added to `background_shells`; indexer stores the path from the spawn tool_result ("Output is being written to: ...")
- `POST /shells/{tool_use_id}/kill`: uses `lsof` to find the PID(s) writing to the output file, sends `SIGTERM`, always marks `terminated_by='kill'` in the DB (safe no-op if the process already exited)
- SvelteKit proxy at `/api/shells/[tool_use_id]/kill` routes the client POST to FastAPI (avoids CORS)
- Button wired: shows "…killing" while in-flight, auto-refreshes list on completion

---

### ✅ Current State
- All CI checks passing (Python lint, format, tests)
- `/shells` page fully functional: list renders, all filters work, Refresh works, session links work
- Live shells show RUNNING with green pulse; naturally-exited shells flip to CLOSED at the exact exit timestamp
- Kill Shell button terminates the process and immediately updates the row
- Poll output logs display for all sessions using modern Claude Code

the-non-expert:
## Cron fires + UI fixes (this commit)

### Why fires weren't showing on `/cron`

`GET /cron` called `get_cron_global()` which was a pure SQL join — it never touched any JSONL files. Fire inference (`infer_cron_fires_bulk`) requires reading each session's JSONL, so it only existed on the per-session endpoint. The global page explicitly said _"Fire history is only available in the per-session view"_.

**Key insight**: the session JSONL is already on disk whether the session is live or closed. The 7-day TTL means a cron job's source session is always available to read. No reason not to do it globally.

### What changed

**API (`api/db/queries_shells_cron.py`, `api/routers/cron.py`)**
- `get_cron_global()` now accepts `include_fires=True` + `claude_projects_dir`
- Groups rows by `session_uuid`, reads each JSONL exactly once, runs `infer_cron_fires_bulk()` per session, attaches `fires` + `latest_state` to every job — identical to the per-session path
- Router exposes `?include_fires=true` (default on)

**Frontend fixes**
- Removed the "fire history only in session view" fallback copy — fires now populate on the global page
- **Project filter bug fixed**: dropdown options now carry `{encoded, label}` pairs; the URL param and server filter both use the encoded name; client-side filter compares `project_encoded_name` (was comparing display name against encoded name → server filter silently returned all jobs)
- `ConversationView` now passes `projectEncodedName` to `SessionCronSection` so the "View all in project →" link actually works
- Fires list capped at `max-height: 400px` with inline scroll in both `/cron` and session cron tab
- `/cron` expanded panel restructured: metadata fields wrap in a strip above, prompt block and fires scroll sit side-by-side below so they share the same natural row height

the-non-expert:
## What's new in this push

### Bug fixes
- **Manual shells now appear immediately** — `GET /shells` triggers an incremental JSONL re-index (debounced to max once per 10s) before reading from the DB. Previously shells spawned via `! <cmd>` in the Claude Code CLI could take up to 5 minutes to appear because the periodic reindex runs every 300s. Now they show up on the next refresh cycle.
- **User-initiated shell exit detection** — `_parse_task_notification` previously required `<tool-use-id>` to be present; user-initiated shells omit it. Fixed to fall back to `<task-id>` lookup via `shell_id_to_tu` so natural exits are correctly recorded for manual shells too.

### New features
- **Auto-refresh** — The shells page now polls `invalidateAll()` every 5 seconds automatically. No need to manually click ↻.
- **Live "Updated Xs ago" footer** — The hardcoded `"Updated 2s ago"` string is now a real counter that resets on each refresh cycle.

### UI polish (poll cards)
- **Human-readable timestamps** — Raw ISO strings like `2026-05-15T07:04:17.120Z` are replaced with `"May 15 at 7:04 AM"` using the existing `formatSpawned` formatter.
- **Bytes beside copy button** — Size indicator moved from a separate absolute-positioned overlay (which overlapped the scrollbar) into the poll header row, inline with the copy button. No extra height, no overlap.
- **Collapsed card cleanup** — Bytes count removed from the collapsed card summary; it's redundant once you expand the poll detail.
- **Minimal scrollbar** — Custom webkit + Firefox scrollbar styles on poll `<pre>` blocks: 3px wide, nearly invisible track, faint thumb (`rgba(0,0,0,0.12)`) that darkens slightly on hover — enough to signal scrollability without visual noise.

JayantDevkar:
## Review pass + merge-prep fixes (pushed in `4fc3e70`, `d260c5c`)

Ran a focused multi-agent review across the whole diff (DB/data layer, API routers + services + hooks, the #76 token/cost math, frontend, and test quality). Headline: solid PR, no security or data-loss landmines, clean merge. One real functional bug and a couple of portability/correctness issues were found and fixed; everything else is deferred to a follow-up.

### Fixed in this push
- **🔴 HIGH — killed shells reverted to RUNNING.** The shell UPSERT overwrote `terminated_at`/`terminated_by`/`exit_code` with values re-derived from the JSONL on every re-index. A karma-side kill never appears in the transcript, so the re-derivation produced `NULL` and clobbered the kill stamp — and since `GET /shells` triggers a debounced re-index (≤10 s), the kill button **visibly self-reverted within seconds**. Termination fields now `COALESCE` existing-wins (sticky once closed; first-time `NULL→set` natural-exit detection still works); `output_file_path` coalesces fresh-wins with a fallback to the stored path.
- **🟠 MED — missing `ShellToolName.MANUAL`.** The indexer writes `tool_name='Manual'` for `!`-backgrounded shells (allowed by the v19 CHECK), so `BackgroundShell.model_validate()` on such a row would have raised. Added the enum member.
- **🟠 MED — `/cron` "Rescan logs" hardcoded `http://localhost:8000`.** Broke under any non-local `PUBLIC_API_URL`. Now uses the shared `API_BASE`.
- **🟡 LOW** — corrected the `kill_shell` docstring (no SIGKILL fallback actually exists), consistent ISO timestamp in the kill path, thread-safe `_maybe_sync()` debounce (lock the check-and-set), and a top-level catch-all in `live_session_tracker.main()` so a hook failure can never exit non-zero / surface as a Claude Code hook error.

Full API suite green locally (1217 + 454 passed) and **all 20 CI checks pass** on the push. `ruff check`/`format` clean on the CI-scoped dirs.

### PR description corrected
- **Bug 1** — the description had it backwards. The long-context >200K threshold **correctly includes cache-read tokens** (matches Anthropic billing; the model attends to them), as the code comment and `test_long_context_includes_cache_in_threshold` confirm. Commit `7a01dee` was a comment-only clarification, not a behavior change. Description updated.
- **Bug 6** — cache reads are now **folded into** Total Tokens (previously excluded); they are not yet surfaced as a *separate* stat. Description updated to match.

### Verified solid (no action needed)
Schema chain (v13/v18/v19, `SCHEMA_VERSION=19`) idempotent + cross-branch-safe; no SQL injection (all dynamic `IN()` parameterized); kill endpoint safe for a localhost tool (request supplies only `tool_use_id` bound as a param; `output_file_path` is DB/indexer-sourced; argv, no `shell=True`); `live_session_store` shallow-merge not exploitable today; token-cost bugs 2–5 correct; frontend has no XSS and no interval leaks; the flagged "error-count off-by-one" was a false alarm.

### Deferred → #78
Test-coverage gaps for the highest-risk new paths (`_is_session_ended`, the kill endpoint, `<task-notification>` parsing, the modern Read-poll path, Manual extraction, the v19 migration), a dev-machine test-isolation hang in the shells router fixture, and three minor nits (static `/cron` TTL countdown, `/shells` project-filter URL desync, `get_primary_model()` non-determinism).

PR #77, opened 2026-06-04
fix(token-cost): fix 6 token/cost calculation bugs

### LINKED ISSUE #76: fix(token-cost): 6 bugs in token/cost calculation and display

the-non-expert:
Closes #76.

## Summary

- **Bug 3** `analytics.py:79` — Cache hit rate denominator double-counted `cache_creation` tokens (already inside `total_input`). One-liner fix.
- **Bug 2** `indexer.py:495` + `agent.py` — Subagent cost was calculated with the default Sonnet pricing regardless of the actual model. Added `get_models_used()` / `get_primary_model()` to `Agent` and pass the model through to `calculate_cost()`.
- **Bug 4** `analytics.py:546` — Project analytics split tokens evenly across models with integer division, then re-priced each split. Replaced with `session.get_total_cost()` which already sums per-message costs at the correct per-message model.
- **Bug 5** `plugins.py:606` — Plugin cost attribution hardcoded Sonnet prices (`$3/M`, `$15/M`) and ignored cache tokens. Replaced with `session.get_total_cost()`.
- **Bug 1** `usage.py:189` — Documented why the long-context threshold correctly uses total context including cache reads (model attends to all tokens, per Anthropic billing). No behavior change.
- **Bug 6** `analytics.py:92` — "Total Tokens" stat excluded cache-read tokens. Now includes `total_cache_read` in the display total on both DB fast path and full-parse fallback.

## UI disclaimer

Added a `footnote` prop to `StatsCard`/`StatsGrid`/`StatItem` and wired a *"Pay-as-you-go API rate — not your subscription cost"* note onto every cost stat card: session view, project analytics, and global analytics.

## Test plan

- [x] `pytest tests/` — 1587 passed, 6 skipped, 0 failures
- [x] `npm run check` — 0 errors, 10 pre-existing warnings
- [x] Both API (:8000) and frontend (:5173) servers running

### DIFF
--- a/api/db/indexer.py
+++ b/api/db/indexer.py
@@ -492,7 +492,7 @@ def _index_session(
                         subagent_type,
                         usage.total_input,
                         usage.output_tokens,
-                        usage.calculate_cost(),
+                        usage.calculate_cost(subagent.get_primary_model()),
                         duration,
                         subagent.start_time.isoformat() if subagent.start_time else None,
                     ),
--- a/api/models/agent.py
+++ b/api/models/agent.py
@@ -53,6 +53,7 @@ class AgentCache(BaseCache):
         "tasks",
         "skills_used",
         "commands_used",
+        "models_used",
     )
 
     def __init__(self) -> None:
@@ -64,6 +65,7 @@ class AgentCache(BaseCache):
         self.tasks: Optional[List["Task"]] = None
         self.skills_used: Optional[Dict[tuple, int]] = None  # {(name, source): count}
         self.commands_used: Optional[Dict[tuple, int]] = None  # {(name, source): count}
+        self.models_used: Optional[set] = None
 
     def reset(self) -> None:
         """Reset all cache values to initial state."""
@@ -74,6 +76,7 @@ class AgentCache(BaseCache):
         self.tasks = None
         self.skills_used = None
         self.commands_used = None
+        self.models_used = None
         super().reset()
 
 
@@ -216,6 +219,7 @@ class Agent(BaseModel):
         last_ts: Optional[datetime] = None
         usage = TokenUsage.zero()
         message_count = 0
+        models: set = set()
         # Skills/commands use (name, source) tuple keys for invocation tracking
         skills: Counter = Counter()  # {(name, source): count}
         commands: Counter = Counter()  # {(name, source): count}
@@ -232,6 +236,8 @@ class Agent(BaseModel):
             if isinstance(msg, AssistantMessage):
                 if msg.usage:
                     usage = usage + msg.usage
+                if msg.model:
+                    models.add(msg.model)
 
                 # Extract skills/commands from Skill tool uses
                 for block in msg.content_blocks:
@@ -249,6 +255,7 @@ class Agent(BaseModel):
         cache.end_time = last_ts
         cache.usage_summary = usage
         cache.message_count = message_count
+        cache.models_used = models
         cache.skills_used = dict(skills)
         cache.commands_used = dict(commands)
         cache.mark_loaded(self._get_file_mtime())
@@ -332,6 +339,15 @@ class Agent(BaseModel):
         """
         self._load_metadata()
         return self._get_cache().commands_used or {}
+
+    def get_models_used(self) -> set:
+        """Get set of all models seen in assistant messages (cached)."""
+        self._load_metadata()
+        return self._get_cache().models_used or set()
+
+    def get_primary_model(self) -> Optional[str]:
+        """Return the first model encountered, or None if no messages have a model."""
+        return next(iter(self.get_models_used()), None)
 
     @property
     def start_time(self) -> Optional[datetime]:
--- a/api/models/usage.py
+++ b/api/models/usage.py
@@ -186,6 +186,9 @@ class TokenUsage(BaseModel):
         """
         pricing = _resolve_model(model)
 
+        # Anthropic's long-context surcharge applies to total context length, which includes
+        # cache-read tokens — the model still attends to them. Using total context here is
+        # correct per billing; the high numbers come from legitimate large-context sessions.
         total_input = (
             self.input_tokens + self.cache_creation_input_tokens + self.cache_read_input_tokens
         )
--- a/api/routers/analytics.py
+++ b/api/routers/analytics.py
@@ -30,8 +30,6 @@ router = APIRouter()
 logger = logging.getLogger(__name__)
 
 
-
-
 def _get_analytics_sqlite(
     project: Optional[str],
     start_dt: Optional[datetime],
@@ -86,7 +84,9 @@ def _get_analytics_sqlite(
 
         return ProjectAnalytics(
             total_sessions=totals["total_sessions"],
-            total_tokens=totals["total_input"] + totals["total_cache_read"] + totals["total_output"],
+            total_tokens=totals["total_input"]
+            + totals["total_cache_read"]
+            + totals["total_output"],
             total_input_tokens=totals["total_input"],
             total_output_tokens=totals["total_output"],
             total_duration_seconds=totals["total_duration"],
--- a/api/routers/analytics.py
+++ b/api/routers/analytics.py
@@ -30,7 +30,6 @@ router = APIRouter()
 logger = logging.getLogger(__name__)
 
 
-from models.usage import TokenUsage
 
 
 def _get_analytics_sqlite(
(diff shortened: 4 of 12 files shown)


the-non-expert:
Folded into #75 (feat/session-id-resume-and-live-tracking-cleanup) via rebase. All token-cost fixes are now included in that PR.

PR #86, opened 2026-07-18
feat(live-sessions): "open terminal" button to focus a session's terminal window

JayantDevkar:
## Summary

Adds a button on the live session page that raises (focuses) the terminal window/pane the Claude Code session is running in.

The literal request — "take me to the terminal window" — can't be done by a browser link alone (a web page can't focus another app's OS window). But Karma runs **locally**: the API is on the same machine as the terminals. So the browser button calls the local API, which shells out to the platform's window manager. Two things had to be built, neither of which existed before: **capturing** which terminal a session runs in, and **acting** on it from the API.

On macOS the focus is **exact-window**: the captured pid (the live `claude` process) is resolved to its tty at click time and AppleScript selects the precise Terminal.app/iTerm2 window+tab owning that tty — not just the app.

Design & spec: `docs/feature/open-terminal/spec.md` (included in this PR).

## Changes

- **hooks/live_session_tracker.py** — `resolve_terminal()` captures terminal identity at SessionStart (`TMUX`/`TMUX_PANE`, `TERM_PROGRAM`/`TERM_SESSION_ID`, `WINDOWID`, and the `claude` pid); any other hook event **backfills** it when missing, so pre-existing sessions become focusable on their next activity.
- **api/models/live_session.py** — new `TerminalInfo` model + `LiveSessionState.terminal` field.
- **api/schemas.py** — `TerminalInfo`, `TerminalFocusResult`, and `terminal` + computed `can_focus_terminal` on `LiveSessionSummary`.
- **api/services/terminal_focus.py** — best-effort, never-raises focus service: tmux (any OS); macOS **exact tab** for Terminal.app/iTerm2 via pid → `ps -o tty=` → AppleScript tty match, falling back to app activation (`osascript`) for other terminals or on any miss; `xdotool`/`wmctrl` (Linux X11).
- **api/routers/live_sessions.py** — `POST /live-sessions/{id}/focus-terminal`; threads terminal fields through `state_to_summary`.
- **frontend** — shared `TerminalFocusButton.svelte` (label + icon variants): labeled button in `ConversationHeader`, subtle icon button on the home live-sessions strip and the `/sessions` live section rows. List rows use a stretched-link pattern (button is a sibling of the row `<a>`, not nested), success/failure is announced via `aria-live` and conveyed by glyph shape (check/X) not color alone, and the POST is aborted on unmount / after 10s. Shown only when `can_focus_terminal` is true; new TS types. Reviewed by designer + code-review agents; all findings applied.
[...]

PR #92, opened 2026-07-21
fix(terminal-focus): reliable Terminal.app raise — retry + out-of-process verify, no miniaturize, honest cross-Space reporting

JayantDevkar:
## What happened

The "open terminal" button (#86/#87/#89) degraded from working in the morning to intermittent by afternoon (2026-07-21). Live systematic debugging against the running dashboard reproduced the failure through a real browser click and isolated three measured defects in the Apple_Terminal path of `terminal_focus.py` — all in the AppleScript layer, none in the #89 identity/lifecycle logic:

1. **Reorder-churn race** — `set index of window id X to 1` + `activate` genuinely fails ~40% of the time (measured 2/5) when it closely follows other window reordering, e.g. the user clicking the browser right before the button. The window server needs ~0.3–0.5s to settle.
2. **Stale in-script verification** — reading `id of front window` inside the same AppleScript run that mutated the order returned pre-mutation state in both directions during trials: false "not raised" *and* one false "raised" (fresh-process ground truth disagreed).
3. **Harmful miniaturize fallback** — on a verification miss, the miniaturize/un-miniaturize cycle restored the window to the *back* of the z-order. Observed live: the target went from front to index 7 (bottom), so a transient miss both reported failure and buried the user's window, making the next click start from a worse state.

Plus a fourth mechanism explaining the time-of-day pattern: **`activate` never switches macOS Spaces while Terminal has a window on the current Space**, so a raise can top the z-order while staying invisible to the user. As windows spread across Spaces during the day, more sessions silently fell into this mode (5 of 7 Terminal windows were off-Space when debugged).

## What this PR changes

- The Apple_Terminal raise script now *only* selects the tab, sets `index to 1`, and activates, returning the matched window id — no in-script verification, no miniaturize cycle, so a transient miss can no longer bury the target window.
- New `_focus_terminal_app_tab()` drives the raise from Python: up to 3 attempts with 0.15s settle each, verifying via a **separate** `osascript` process (`_TERMINAL_VERIFY_SCRIPT`), so verification reads are never stale-in-transaction.
- New cross-Space probe `_window_on_current_space()`: System Events enumerates only current-Space windows; if the raised window's title isn't there, the API returns `focused=false` with "the window is on another desktop or full screen — find it via Mission Control" instead of a false success. Titles are sanitized to stable ASCII (`_sanitize_title`) so Terminal's animated spinner glyphs and unicode dashes don't break the match between two reads.
- Probe failure (no Accessibility permission for System Events) degrades to trusting z-order — machines without that permission keep exactly the old success behavior, never a new failure.
- Exhausted retries return an honest `focused=false` ("could not be raised after 3 attempts — try clicking again") and never fall through to blind app activation.
- iTerm2, tmux, and Linux paths are unchanged; the tmux GUI-raise path picks up the same hardened Terminal.app mechanics automatically.

## How it was verified

- **TDD**: 7 new behavioral tests (separate-process verify, retry-on-stale-front, honest give-up, cross-Space honesty, permission fallback, glyph-tolerant matching, structural guard that the raise script contains no `miniaturized`/`front window`). The 5 core ones were watched failing against the old code for the right reasons before implementation. Focus suite: **53/53 pass**; ruff clean.
- **Full API suite**: 1771 passed. 6 failures are pre-existing and unrelated — `tests/api/test_background_shells_router.py` leaks real machine state (the endpoint reads the live `~/.claude_karma/metadata.db` instead of the test fixture DB; verified by rerunning one test and seeing real shell rows in the "empty DB" assertion).
- **Live, API layer**: under a churn harness (bury target at z-bottom, immediately POST `/focus-terminal`), the old code failed 2/5 one-shot raises; the fixed endpoint raised **5/5**, each confirmed against window-server ground truth from a fresh osascript process — not the API's own claim.
- **Live, full stack**: **3/3** real Playwright clicks on the dashboard button with the target window deliberately buried first — all returned `focused=true` and ground truth showed the window front at index 1. This is the exact browser-click scenario that failed before the fix.
- Both cross-Space clicks during testing reported truthfully (macOS genuinely switched Spaces in those runs and the window was verifiably visible after).

## Not in this PR (follow-ups)

- **Cross-Space refusal path live repro**: unit-tested but not staged live — the active Space kept legitimately changing under the user during testing, and when the raise lands, `focused=true` is correct. If the "another desktop" tooltip appears in real use, that's this path working as designed.
- **Background-shells test isolation**: the 6 pre-existing `test_background_shells_router.py` failures (real DB leaking into fixtures) need their own fix/ticket — untouched here.
- **Frontend**: no changes needed — `TerminalFocusButton.svelte` already surfaces `detail` as the button tooltip, so the new honest messages reach the user as-is.

https://claude.ai/code/session_01UVoojXY7ZuQc9LHBg76Ueb

---

## Component

- [x] API (FastAPI backend)

## Type

- [x] Bug fix

## Testing

- [x] API tests pass (`cd api && pytest`) — 53/53 focus suite; full suite 1771 passed, 6 pre-existing unrelated failures (see body)
- [x] Manually tested in browser — 3/3 Playwright churn trials, ground-truth verified

## Related Issues

Relates to #86, #87, #89 (terminal-focus series)

### DIFF
--- a/api/services/terminal_focus.py
+++ b/api/services/terminal_focus.py
@@ -16,6 +16,15 @@ which stores the claude process's tty (and iTerm session id / app bundle id)
 *while the process is alive*. Focus prefers those stored identifiers over a
 click-time pid lookup, so tab matching survives pid death and recycling.
 
+Terminal.app raising is racy: ``set index to 1`` genuinely fails ~40% of the
+time right after other window reordering (the window server needs ~0.3–0.5s
+to settle), and window-order reads *inside the mutating script* return stale
+values in both directions. The raise is therefore retried with settle
+delays and verified from a separate ``osascript`` process. ``activate`` also
+never switches macOS Spaces when the app already has a window on the current
+Space — a cross-Space window can top the z-order while staying invisible, so
+Space visibility is probed via System Events and reported honestly.
+
 Everything is best-effort and honest: if a window genuinely could not be
 raised, the result says ``focused=False`` with a human-readable reason
 instead of raising — or worse, raising the wrong window.
@@ -24,6 +33,7 @@ instead of raising — or worse, raising the wrong window.
 from __future__ import annotations
 
 import os
+import re
 import shutil
 import subprocess
 import sys
@@ -263,12 +273,12 @@ _TAB_SCRIPT_APPS: Dict[str, str] = {
 }
 
 # Per-app AppleScript to select the exact tab owning a tty. `{tty}` is
-# substituted with e.g. /dev/ttys006. Terminal.app ignores `set frontmost`,
-# but `set index to 1` reorders reliably (windows parked on other Spaces
-# silently ignore `set miniaturized`, which is why index comes first and the
-# miniaturize cycle is only a fallback). The outcome is VERIFIED: the script
-# reports "raised" only when the matched window really is frontmost after
-# the attempts; anything else reports "selected" so the API stays honest.
+# substituted with e.g. /dev/ttys006. Terminal.app's script only selects and
+# raises (`set index to 1` + activate) and returns the matched window id —
+# NO in-script verification (order reads are stale within the mutating
+# script) and NO miniaturize fallback (programmatic un-miniaturize restores
+# the window to the BACK of the z-order, burying what was just raised).
+# Verification happens in Python via a separate osascript process.
 _MAC_TAB_SCRIPTS: Dict[str, str] = {
     "Apple_Terminal": (
         'tell application "Terminal"\n'
@@ -288,19 +298,7 @@ _MAC_TAB_SCRIPTS: Dict[str, str] = {
         "        set index of window id matchedId to 1\n"
         "    end try\n"
         "    activate\n"
-        "    if id of front window is not matchedId then\n"
-        "        try\n"
-        "            set miniaturized of window id matchedId to true\n"
-        "            delay 0.4\n"
-        "            set miniaturized of window id matchedId to false\n"
-        "            delay 0.2\n"
-        "        end try\n"
-        "    end if\n"
-        "    if id of front window is matchedId then\n"
-        '        return (matchedId as text) & " raised"\n'
-        "    else\n"
-        '        return (matchedId as text) & " selected"\n'
-        "    end if\n"
+        "    return matchedId as text\n"
         "end tell"
     ),
     "iTerm.app": (
@@ -322,6 +320,34 @@ _MAC_TAB_SCRIPTS: Dict[str, str] = {
         "end tell"
     ),
 }
+
+# Raise verification, run as a SEPARATE osascript process — reads from the
+# script that mutated the order are stale (measured both false negatives and
+# false positives live). Returns "<front id>\n<target window name>".
+_TERMINAL_VERIFY_SCRIPT = (
+    'tell application "Terminal"\n'
+    "    return (id of front window as text) & linefeed & (name of window id {window_id})\n"
+    "end tell"
+)
+
+# System Events only enumerates windows on the CURRENT Space — exactly the
+# probe needed to tell "raised where the user can see it" from "topped the
+# z-order on another desktop". Requires Accessibility permission; failure is
+# treated as unknown and z-order is trusted.
+_SE_WINDOW_NAMES_SCRIPT = (
+    'tell application "System Events" to tell process "Terminal"\n'
+    "    set out to \"\"\n"
+    "    repeat with w in windows\n"
+    "        set out to out & (name of w) & linefeed\n"
+    "    end repeat\n"
+    "    return out\n"
+    "end tell"
+)
+
+# Post-churn, the first raise reliably fails; ~0.45s of settle almost always
+# lands it (measured 9/10, nearly all on the third attempt).
+_RAISE_ATTEMPTS = 3
+_RAISE_SETTLE_SECONDS = 0.15
 
 # Same shape, but matches an iTerm2 session by its unique id (the UUID from
 # ITERM_SESSION_ID). Survives pid death and recycling entirely.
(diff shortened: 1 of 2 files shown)


api/services/terminal_focus.py
```diff
@@ -288,19 +298,7 @@ def _focus_tmux(pane: str) -> Dict[str, Any]:
...
-        "            set miniaturized of window id matchedId to false\n"
-        "            delay 0.2\n"
-        "        end try\n"
-        "    end if\n"
-        "    if id of front window is matchedId then\n"
-        '        return (matchedId as text) & " raised"\n'
-        "    else\n"
-        '        return (matchedId as text) & " selected"\n'
-        "    end if\n"
+        "    return matchedId as text\n"
```
JayantDevkar, line 301: The raise script deliberately returns only the matched window id — the old script's in-script verification (`id of front window`) read stale state within the same Apple Event transaction (measured both directions: false "not raised" and one false "raised" that fresh-process ground truth contradicted), and its miniaturize fallback restored the window to the BACK of the z-order, burying the window it had just raised (observed live: target demoted from front to index 7). Verification moved to Python via a separate osascript process; the miniaturize cycle is gone entirely.

api/services/terminal_focus.py
```diff
@@ -323,6 +321,34 @@ def _focus_tmux(pane: str) -> Dict[str, Any]:
...
+    "    repeat with w in windows\n"
+    "        set out to out & (name of w) & linefeed\n"
+    "    end repeat\n"
+    "    return out\n"
+    "end tell"
+)
+
+# Post-churn, the first raise reliably fails; ~0.45s of settle almost always
+# lands it (measured 9/10, nearly all on the third attempt).
+_RAISE_ATTEMPTS = 3
```
JayantDevkar, line 360: Why 3 × 0.15s: under a reorder-churn harness (bury target, immediately raise), the one-shot raise failed 2/5; with this retry loop it recovered 9/10, with success almost always on attempt 3 — i.e. ~0.3–0.45s of total settle is what the window server actually needs. Fewer attempts measurably miss; more just adds latency to genuine failures. Worst-case happy path cost is one 150ms sleep before the first verify.

api/services/terminal_focus.py
```diff
@@ -323,6 +321,34 @@ def _focus_tmux(pane: str) -> Dict[str, Any]:
...
+    'tell application "Terminal"\n'
+    "    return (id of front window as text) & linefeed & (name of window id {window_id})\n"
+    "end tell"
+)
+
+# System Events only enumerates windows on the CURRENT Space — exactly the
+# probe needed to tell "raised where the user can see it" from "topped the
+# z-order on another desktop". Requires Accessibility permission; failure is
+# treated as unknown and z-order is trusted.
+_SE_WINDOW_NAMES_SCRIPT = (
```
JayantDevkar, line 345: System Events enumerates only windows on the user's CURRENT Space — that asymmetry (Terminal's own AppleScript sees all 7 windows, SE saw 2) is exactly the probe needed to distinguish "raised where the user can see it" from "topped the z-order on an invisible desktop". `activate` does not switch Spaces while Terminal has any window on the current Space, so without this probe the API reports success the user never sees — the main mechanism behind "worked in the morning, unreliable by afternoon" as windows spread across Spaces.

api/services/terminal_focus.py
```diff
@@ -414,6 +440,82 @@ def _parse_tab_result(
     }
 
 
+def _sanitize_title(title: str) -> str:
+    """Reduce a window title to stable ASCII (spinner glyphs animate between reads)."""
+    return re.sub(r"\s+", " ", re.sub(r"[^ -~]", "", title)).strip()
```
JayantDevkar, line 456: Why strip to printable ASCII: real Terminal titles look like `claude-karma — ⠐ Debug terminal session… — 142×43` — the braille spinner glyph animates between the Terminal read and the System Events read, so exact matching fails on live sessions (an AXRaise-by-name experiment failed 8/8 on precisely this). The em-dash and × also differ between the two APIs' renderings; ASCII-only comparison survives all of it. Collision risk (two identically-titled same-size windows) only affects which *message* is returned, never which window gets raised — matching is still by tty.

api/services/terminal_focus.py
```diff
@@ -414,6 +440,82 @@ def _parse_tab_result(
...
+    """Whether a window with this title is on the user's current Space; None = unknown."""
+    target = _sanitize_title(target_name)
+    if not target:
+        return None
+    try:
+        res = _run(["osascript", "-e", _SE_WINDOW_NAMES_SCRIPT])
+    except (subprocess.SubprocessError, OSError):
+        return None
+    if res.returncode != 0:
+        return None  # likely no Accessibility permission
```
JayantDevkar, line 479: Permission fallback consequence: if the API process lacks Accessibility permission for System Events, the Space probe returns None (unknown) and the z-order verification alone decides — which is exactly the pre-PR success criterion. So machines without the permission keep the old behavior; the new focused=false path can only trigger on a *confirmed* off-Space window, never on a probe failure.

api/services/terminal_focus.py
```diff
@@ -414,6 +440,82 @@ def _parse_tab_result(
...
+            return None  # no tab owns this tty — let the caller fall back
+        time.sleep(_RAISE_SETTLE_SECONDS)
+        front, name = _terminal_front_and_name(window_id)
+        if front != window_id:
+            continue  # settle race — reorder didn't take yet; try again
+        if _window_on_current_space(name) is False:
+            return {
+                "focused": False,
+                "method": "osascript-tab",
+                "detail": (
```
JayantDevkar, line 542: Why the cross-Space case returns immediately instead of retrying: front == window_id proves the raise itself took — retrying `set index to 1` cannot move a window between Spaces, so further attempts would just burn 300ms and return the same answer. Only the front-mismatch (settle race) case loops.

api/tests/test_terminal_focus.py
```diff
@@ -41,8 +41,16 @@ def _fake_run_factory(
...
+    front_ids=None,
 ):
-    """_run stub routing ps / is-running probes / tab scripts / activation."""
+    """_run stub routing ps / is-running probes / raise / verify / SE scripts."""
+    window_id = tab_stdout.split()[0] if tab_stdout.split() else ""
+    names = se_names if se_names is not None else (target_name,)
+    fronts = list(front_ids) if front_ids is not None else [window_id]
+    state = {"verify": 0}
 
     def fake_run(cmd):
```
JayantDevkar, line 61: Factory contract change for reviewers of the test diff: scripts are now routed by distinguishing content — `sessions of t` (iTerm2 one-shots, unchanged full "<id> raised" output), `tty of t is` (Terminal raise, bare id), `front window` (verify, "<front id>\n<title>"), `System Events` (current-Space titles). The `front_ids` sequence parameter is what lets tests script a stale front on attempts 1–2 and a landed raise on attempt 3.

JayantDevkar (commented):
Inline context for reviewers — one note per change area. All numbers below come from live measurement against the running dashboard during the debugging session that produced this PR.

the-non-expert (approved):

JayantDevkar:
## Review response + re-verification (head `68c1459`)

An internal review pass flagged a real gap in the Space-visibility probe: **`_window_on_current_space` matched windows by title alone**, and Terminal titles are not unique (every uncustomized idle tab is `-zsh — 80×24`). A same-titled window on the current Space could therefore fake visibility for a target sitting on another Space — a false `focused=true` in exactly the case the probe exists to report honestly.

**Done in `68c1459`.** Before choosing the fix I checked the reviewer's suggested alternative on the live machine: System Events exposes **no window id** (AX properties are role/title/position/size/subrole only), so exact-id matching is impossible. The probe now requires the SE `position`+`size` to equal the `bounds` Terminal reports for the exact target window id. Coordinate-system agreement was verified live before implementing: SE position = `x1,y1`, SE size = `x2−x1,y2−y1` (`0,69,1728,1117` ↔ `0,69 / 1728×1048` for the same window). A residual collision now requires a same-titled window at *identical position and size* on the current Space — and the geometry requirement also inertly rejects the phantom `AXUnknown` menu-bar-strip windows (empty title, `1728×33`) that System Events lists for Terminal.

Two smaller review notes, both accepted as-is:
- **Worst-case latency** (~35s if Terminal beachballs through 3×(raise+verify)+probe, each capped at 5s): accepted — the route is sync-def so it runs in FastAPI's thread pool and can't block other requests, and the frontend button already aborts its fetch at 10s. Normal-path cost is unchanged (~0.5s).
- The raise-script/verify-script split, escaping, and detail-string honesty all came back clean in the same review.

### Re-verification at `68c1459`
- Focus suite **55/55** (two new tests: same-title-different-geometry must report "another desktop"; phantom SE windows must never satisfy the probe). A mutation check — temporarily reverting the probe to title-only — made the new collision test fail, confirming it guards the regression it was written for.
- `ruff check` + `ruff format --check` clean (rebased onto @the-non-expert's `ef77a29` CI-format fix — thanks — and ran the same formatter on the new code).
- Live against the running dashboard post-rebase: `POST /focus-terminal` returned `focused=true` via an **earned** geometry+title match (verified the SE line `0,69,1728,1048|<title>` matches the Terminal-side bounds for the same window id — not the unknown-fallback path).

api/services/terminal_focus.py
JayantDevkar: Update at 68c1459: the collision caveat in this note is now stronger than written — after the review finding, the Space probe requires exact geometry (SE position/size == Terminal bounds for the target window id) in addition to the sanitized title, so a same-titled window no longer even changes which *message* is returned unless it also sits at identical position and size. Title sanitization itself is unchanged.

api/services/terminal_focus.py
JayantDevkar: Update at 68c1459: the probe emits "x,y,w,h|name" per window now, not bare names — visibility requires geometry + title to match (titles are not unique; review finding). Also verified live that System Events exposes no window id (AX properties are role/title/position/size only), which is why geometry is the disambiguator rather than an exact id match. Coordinate agreement verified live: SE position = x1,y1, size = x2−x1,y2−y1 of Terminal bounds.

JayantDevkar:
## Test-review response (head `167f7629`)

A second internal review pass (test-suite lens) came back **approve**: it independently confirmed all five pinned regression scenarios (in-script verify reintroduction, attempt count in both directions, cross-Space no-retry, SE-failure z-order trust, empty-raise fallback), checked every real script constant against the fake router for silent-swallow routing (none), and traced the three pre-existing tests touched by the factory rewrite (none weakened).

It flagged three non-blocking coverage gaps — all real, so they're closed rather than deferred. **Done in `167f7629`** (4 tests, no production changes):

- **Settle delay was never spied** — the sharpest one: deleting `time.sleep(_RAISE_SETTLE_SECONDS)` passed the entire suite while silently reintroducing the ~40% churn race this PR exists to fix. Now spied and wired to the constant; mutation-checked (removing the sleep fails the new test).
- **Truncated verify output** (front id line only) → visibility unknown → trust z-order. Pinned as `focused=true`.
- **Verify script erroring** (window closed between raise and verify, `Invalid index (-1719)`) → counts as not-raised, retries ×3, honest "could not be raised". Pinned.
- **Mid-loop `SubprocessError` from the raise script** → abandons the exact-tab path for the guarded app-activation fallback, never crashes or claims an `osascript-tab` success. Pinned.

Suite at head: **59/59**, `ruff check` + `ruff format --check` clean. No production code changed in this commit, so the live verification posted for `68c1459` still covers the head's runtime behavior.

PR #94, opened 2026-07-27
feat(desktop): one-click launcher + autostart for macOS and Windows

the-non-expert:
## What this adds

A desktop icon that starts Karma's servers for you and opens the dashboard, so you no longer need two terminals every time — installable **from the dashboard**, not a shell command.

- **`scripts/karma_desktop/`** — a shared, stdlib-only launcher + installer. The repo path is derived from `__file__` and the Python interpreter is discovered on the target machine, so a clone anywhere works for any user. A test fails the build if any machine-specific path (a home dir, a pinned interpreter) ever reappears in the shipped sources.
- **macOS**: a `.app` bundle, optional Dock pin (reuses the replaced tile's slot), optional launchd login item.
- **Windows**: a Desktop shortcut driven by `pythonw` (no console flash), optional Startup-folder entry. Both shortcuts run through a small boot script kept outside the repo, so a moved/deleted repo shows an alert (or the Startup entry self-removes) instead of a silent dead icon. Taskbar pinning is deliberately not attempted — Windows offers no supported API for it.
- **A native desktop notification** ("Starting servers…") covers the ~1–2 min first `vite` compile so a cold click never looks dead. It replaced an earlier splash window, which needed a throwaway HTTP server and port just to show one line of text.
- **Settings → Desktop App** installs it and toggles autostart. The API endpoint refuses non-loopback and cross-site callers, since it writes an executable bundle and can register a login item; it is localhost-only, not authenticated.
- **Autostart is reversible and independent of the icon**, and its state is read from disk so the dashboard never disagrees with what macOS Login Items / Windows Startup actually show. SETUP.md requires the agent to *ask* before enabling it rather than defaulting it on.

## Supporting changes

- Adopt **5180 / 8020** as Karma's declared ports and pin vite with `strictPort`, so a clash fails loudly instead of silently moving. Sweeps the old 5173/8000 references in code and current docs (historical docs left as-is).
- Ship the **PWA manifest and icons**, and fix a `.gitignore` `*.png` rule that was silently excluding them — the manifest referenced icons git was never tracking.
- CI workflow runs the suite on **ubuntu / windows / macos** × Python 3.10 / 3.12, plus a 3.9 leg on ubuntu to hold the oldest supported interpreter.

## Testing

- **macOS: verified end-to-end on real hardware** — install, Dock pin/position, autostart enable→disable round-trip (launchd load/unload confirmed), cold launch with both servers starting, uninstall, and full cleanup.
- **Windows: covered by CI but not yet hand-run on real hardware** — the `.lnk`/boot-script creation, Startup-folder, shortcut-ownership and path-with-spaces tests only execute on the `windows-latest` runner.
- Linux is explicitly out of scope for now (servers started manually); the installer says so.
- The launcher/installer suite (macOS, Windows and stdlib-only import paths) plus a dedicated localhost-guard suite pass across the CI matrix. The anti-hardcoding guard and the autostart round-trip test were both confirmed to fail against the bugs they guard.

## Notes for the reviewer

- **Manifest overlap with `feat/web-chat`**: that branch also adds `frontend/static/manifest.webmanifest`. Whichever merges second should take one copy — this one uses the correct `--bg-base` theme color and has committed icons.
- The 2 pre-existing `test_background_shells_router` failures on the wider API suite also fail on `main` (they read the real `metadata.db`); unrelated to this change.


scripts/karma_desktop/launcher.py
```diff
@@ -0,0 +1,524 @@
...
+        try:
+            fd = os.open(str(lock), os.O_CREAT | os.O_EXCL | os.O_WRONLY, 0o644)
+            os.close(fd)
+            acquired = True
+        except FileExistsError:
+            import time
+
+            owned = False  # another launch holds the lock: don't spawn
+            try:
+                if time.time() - lock.stat().st_mtime > 60:
```
JayantDevkar: **Bug: the 60s stale threshold contradicts how long this lock is legitimately held.**

The comment at the wait site (below, ~line 489) says the lock is deliberately held across the entire port wait — which on a cold start is the 1–2 minute vite compile, with `WEB_TIMEOUT = 240`. But the lock's mtime is never refreshed during that wait, so any second launch arriving **more than 60 seconds into a legitimate cold start** hits this branch, declares the live lock stale, reclaims it, and spawns a duplicate server set.

That is exactly the "Dock click while launchd's RunAtLoad is still working" race this lock was built to prevent — it misfires in the very scenario it guards. The blast radius is bounded (`strictPort` makes the duplicate vite/uvicorn lose the bind race and exit, so it degrades to log noise), but the guard and its hold duration are internally inconsistent.

Two easy fixes, either works:
1. Touch the lock file's mtime inside the `wait_for_port` polling loop, so a held lock never looks stale; or
2. Raise the stale threshold above `API_TIMEOUT + WEB_TIMEOUT` (~330s), so "stale" can only mean a crashed launch.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/karma_desktop/launcher.py
```diff
@@ -0,0 +1,524 @@
...
+    except OSError:
+        pass
+    try:
+        yield owned
+    finally:
+        if acquired:
+            try:
+                lock.unlink(missing_ok=True)
+            except OSError:
+                pass
```
JayantDevkar, line 410: **Logic flaw (secondary effect of the stale-reclaim bug above): the original winner deletes the reclaimer's lock.**

After a reclaim, the original launch still has `acquired = True` — nothing ever tells it that ownership moved. When its wait finishes, this `finally` unlinks the lock file **by path**, which is now the *reclaimer's* fresh lock, not its own. A third launch arriving in that window then finds no lock at all and spawns yet another server set.

The unlink should only remove a lock this process still owns. Cheapest fix: write the PID into the lock file at creation and only unlink if it still matches (or compare the inode/mtime captured at acquisition). If you go with the mtime-refresh fix on the stale check above, this path becomes unreachable in practice, but the ownership check is still worth having so the two mechanisms can't drift apart again.

---
_Generated by [Claude Code](https://claude.ai/code)_

JayantDevkar (changes requested):
@the-non-expert — thorough in-depth review done (all core files read, both test suites run locally: 37 launcher + 12 API guard tests pass). Overall this is high-quality work: the platform edge-case handling, ownership checks, injection-safe quoting, and the localhost guard are all above the bar, and the user value is real. One bug and one related logic flaw need fixing before merge — both in `_spawn_lock` in `scripts/karma_desktop/launcher.py`, detailed in the two line comments:

1. **Bug — stale-lock reclaim fires during a legitimate cold start.** The lock is held across the full port wait (up to ~330s on a cold vite compile per your own comment at the wait site), but the stale threshold is 60s and the lock's mtime is never refreshed. A second launch >60s into a cold start reclaims the live lock and spawns a duplicate server set — the exact race the lock exists to prevent. Fix: refresh the mtime in the wait loop, or raise the threshold above `API_TIMEOUT + WEB_TIMEOUT`.

2. **Logic flaw — the winner's `finally` unlinks a lock it no longer owns.** After a reclaim, the original launch still has `acquired = True` and deletes the reclaimer's lock by path, reopening the race for a third launch. The unlink needs an ownership check (e.g. PID written into the lock file).

Also, please update the **PR description**: it still describes a splash window "served on an OS-assigned port", but the code replaced that with a desktop notification (`notify()`'s docstring says so explicitly). The merged description shouldn't document a mechanism that no longer exists.

Could you push fixes for the two lock issues and refresh the description? Everything else I found was trivial (stale CORS comment about 5173 in `api/config.py`; the Windows-layout gap in `_is_versioned_interpreter`, e.g. `Python312\python.exe`) and can ride along or land as a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

JayantDevkar (approved):
Re-reviewed `ac44d66` — both findings are properly fixed, verified by reading the diff and running the suite locally (39 passed).

1. **Stale threshold**: `_STALE_LOCK_SECONDS = API_TIMEOUT + WEB_TIMEOUT + 30` sits above the worst-case legitimate hold, so "stale" can now only mean a crashed launch. The reasoning is documented at the constant.
2. **Ownership check**: the lock is stamped with a per-launch `pid:uuid` token and the `finally` only unlinks when the file still carries it — a reclaimed lock is left alone. `FileNotFoundError` on the read falls into the existing `OSError` handler, so no new failure path.
3. The regression tests are exactly right: the converted test pins that a 120s-old *live* lock is not reclaimed (would have failed against the old 60s threshold), and the new token test pins that `finally` never deletes a reclaimed lock.

PR description is also updated to match the notification-based reality. Approving. The two ride-along nits from the previous review (5173 CORS comment, `Python312`-style paths in `_is_versioned_interpreter`) remain optional follow-ups.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #95, opened 2026-07-30
Floating live-sessions dock + Cmd+B session switcher

the-non-expert:
## What this does

Two ways to jump between concurrently-running Claude Code sessions without going back to the sessions list:

- **Floating dock** (bottom-right pill, every page except `/` and `/sessions` where the full table already exists): click to expand a list of live sessions, click a row to open it, or use its dedicated button to focus the session's terminal window.
- **Cmd+B session switcher**: hold Cmd, tap B to cycle a macOS-Cmd+Tab-style HUD through your live sessions, release Cmd to jump to the highlighted one's Timeline tab.

Both read from one shared polling store instead of each hitting `/live-sessions/active` on their own.

## Why Cmd+B, not Cmd+K or Cmd+E

Cmd+K is already the command palette. Cmd+E was tried first but Chrome on Mac reserves it at the browser-chrome level (search-with-selection), so the keydown never reaches page JS — Cmd+B has no such conflict.

## Why the switcher only navigates, doesn't also focus the terminal

Focus-terminal raises the native Terminal window on top of Chrome, which would immediately hide the page you just switched to. Terminal focus stays a separate, deliberate action on the dock's own button.

## Files touched

- `LiveSessionsDock.svelte`, `SessionSwitcher.svelte` — new
- `liveSessionsFeed.ts` — new shared polling store
- `live-session-display.ts` — new shared formatting/URL helpers, extracted out of `LiveSessionsSection.svelte` (also refactored to use them, no behavior change there)
- `+layout.svelte` — mounts both globally
- `KeyboardShortcutsHelp.svelte` — documents Cmd+B

## Test plan

- [x] `npm run check` / `npm run lint` clean (0 errors)
- [x] Verified via curl against the running dev server: homepage and a session-detail route both 200, no SSR errors
- [x] Manual click-through and Cmd+B verification (done live by Ayush during the session)

### DIFF
--- a/frontend/src/lib/components/CommandFooter.svelte
+++ b/frontend/src/lib/components/CommandFooter.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
 	import { onMount } from 'svelte';
-	import { Command, Sun, Moon, Keyboard } from 'lucide-svelte';
+	import { Command, Sun, Moon, Keyboard, ArrowLeftRight } from 'lucide-svelte';
 	import { commandPalette } from '$lib/stores/commandPalette';
+	import { sessionSwitcherStore } from '$lib/stores/sessionSwitcher';
 	import KeyIndicator from '$lib/components/ui/KeyIndicator.svelte';
 
 	interface Props {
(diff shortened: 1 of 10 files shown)


JayantDevkar (approved):
Reviewed via three-agent pass (correctness, simplicity, UI/UX) plus hands-on Playwright verification of every interaction on both macOS and an emulated Windows platform. All blockers found in review were fixed on-branch in a87b1fa5 (cross-platform Ctrl+B, shared-feed migration ending the double-poll, cross-project slug collision in matchesRoute, stray-modifier commit guard) and 4256bd2b (dock under modal scrims, unified pulse semantics, text status labels, AA contrast). Type check, lint, and all CI checks are green; dock, switcher, and LIVE NOW section verified live against real and injected sessions. Remaining follow-ups (HomeSessionsStrip/LiveSessionsTerminal feed migration, HUD focus trap, token-normalizing switcher motion) are non-blocking and documented in the session notes.

PR #96, opened 2026-08-13
Wire timeline search-result highlighting across cards, modal, and tool details

the-non-expert:
## Summary
- Searching a session's timeline found matches but never highlighted them once you opened a result — this wires the search query into every render path that can show one: collapsed card, inline-expanded card, modal popup, `ToolCallDetail`'s per-tool panels, and todo updates.
- Broadens which fields are actually searchable/renderable (Edit `old_string`/`new_string`, Task/Agent `prompt`), with server-side truncation so those additions don't balloon the 1s live-session poll payload.
- Fixes the correctness, performance, and security issues surfaced by two rounds of independent blind code review, including:
  - A per-event search cache that never hit on live sessions (keyed on object identity, which polling replaces every second).
  - A highlight/escape ordering bug where certain queries (`<div>`, `&&`, `"key"`) never highlighted, and others (`lt`, `amp`) corrupted rendered text.
  - A debounce race that could resurrect a search query right after it was cleared via Escape.
  - Untruncated diffs/prompts making highlighting run over multi-MB strings that are clipped to a 320px box anyway.
  - Match-count vs. what's-actually-highlighted divergence, a stuck "hidden match" indicator, whitespace-only-query false positives, and a pre-existing `input.file_path`/`path` key mismatch that left Read/Write/Edit paths blank.

## Test plan
- [x] `npm run check` (frontend) — 0 errors, same 13 pre-existing warnings
- [x] `pytest` (api) — 1795 passed, same 2 pre-existing unrelated failures (local `.storygraph.db` leak, not caused by this branch)
- [x] Both dev servers verified healthy after changes
- [x] Manual verification in-browser (search across collapsed/expanded/modal/tool-detail views, live session behavior)

### DIFF
--- a/api/utils.py
+++ b/api/utils.py
@@ -749,6 +749,18 @@ def find_best_root(path: str, working_dirs: list[str]) -> str | None:
         return None
 
 
+# Cap on free-text tool-input fields (diffs, subagent prompts) placed into
+# timeline metadata. Without this, large Edit diffs or Task prompts ship in
+# full over the timeline endpoint, which live sessions poll every second.
+_METADATA_TEXT_LIMIT = 4000
+
+
+def _cap_text(value: str, limit: int = _METADATA_TEXT_LIMIT) -> str:
+    if len(value) <= limit:
+        return value
+    return value[:limit] + f"\n... [truncated, {len(value)} chars total]"
+
+
 def get_tool_summary(block, working_dirs: list[str] | None = None) -> tuple[str, str | None, dict]:
     """
     Extract title, summary, and metadata from a tool use block.
@@ -780,7 +792,13 @@ def get_tool_summary(block, working_dirs: list[str] | None = None) -> tuple[str,
         return "Write file", to_relative(path), {"path": path, "content": content}
     elif tool_name == "Edit" or tool_name == "StrReplace":
         path = tool_input.get("path") or tool_input.get("file_path", "")
-        return "Edit file", to_relative(path), {"path": path}
+        old_string = _cap_text(tool_input.get("old_string", ""))
+        new_string = _cap_text(tool_input.get("new_string", ""))
+        return (
+            "Edit file",
+            to_relative(path),
+            {"path": path, "old_string": old_string, "new_string": new_string},
+        )
     elif tool_name == "Delete":
         path = tool_input.get("path") or tool_input.get("file_path", "")
         return "Delete file", to_relative(path), {"path": path}
@@ -807,12 +825,14 @@ def get_tool_summary(block, working_dirs: list[str] | None = None) -> tuple[str,
     elif tool_name in ("Task", "Agent"):
         description = tool_input.get("description", "")
         subagent_type = tool_input.get("subagent_type")  # None if missing
+        prompt = _cap_text(tool_input.get("prompt", ""))
         return (
             "Spawn subagent",
             description[:100] if description else None,
             {
                 "description": description,
                 "subagent_type": subagent_type,
+                "prompt": prompt,
             },
         )
     elif tool_name == "TodoWrite":
--- a/api/utils.py
+++ b/api/utils.py
@@ -749,9 +749,7 @@ def find_best_root(path: str, working_dirs: list[str]) -> str | None:
         return None
 
 
-# Cap on free-text tool-input fields (diffs, subagent prompts) placed into
-# timeline metadata. Without this, large Edit diffs or Task prompts ship in
-# full over the timeline endpoint, which live sessions poll every second.
+# Cap on free-text tool-input fields shipped in timeline metadata, which live sessions poll every second.
 _METADATA_TEXT_LIMIT = 4000
 
 
@@ -788,7 +786,7 @@ def get_tool_summary(block, working_dirs: list[str] | None = None) -> tuple[str,
         return "Read file", to_relative(path), {"path": path}
     elif tool_name == "Write":
         path = tool_input.get("path") or tool_input.get("file_path", "")
-        content = tool_input.get("content", "")
+        content = _cap_text(tool_input.get("content", ""))
         return "Write file", to_relative(path), {"path": path, "content": content}
     elif tool_name == "Edit" or tool_name == "StrReplace":
         path = tool_input.get("path") or tool_input.get("file_path", "")
--- a/frontend/src/lib/components/timeline/TimelineEventCard.svelte
+++ b/frontend/src/lib/components/timeline/TimelineEventCard.svelte
@@ -62,14 +62,17 @@
 
 	let isCopied = $state(false);
 
-	// Sanitized markdown for expanded view — parsing/sanitizing is the expensive
-	// part, so this effect depends only on content/expansion state, NOT searchQuery.
-	// Highlighting is applied separately below as a cheap $derived over this.
+	// Sanitized markdown for expanded view; depends on content/expansion, not searchQuery (see $derived below).
 	let sanitizedExpandedContent = $state('');
 
-	// Render markdown when content changes or expansion state changes
+	// Skips tool_call/todo_update — they render via ToolCallDetail/TodoUpdateDetail instead.
 	$effect(() => {
-		if (isExpanded && hasExpandableContent) {
+		if (
+			isExpanded &&
+			hasExpandableContent &&
+			event.event_type !== 'tool_call' &&
+			event.event_type !== 'todo_update'
+		) {
 			const rawContent =
 				event.metadata?.full_content ||
 				event.metadata?.full_thinking ||
@@ -90,12 +93,13 @@
 		}
 	});
 
-	// Re-runs only the (cheap) text-node walk when searchQuery changes, not the
-	// full marked.parse + DOMPurify.sanitize pipeline.
+	// Cheap text-node walk over the already-parsed content; skips tool_call/todo_update like the effect above.
 	const renderedExpandedContent = $derived(
-		searchQuery
-			? highlightHtmlContent(sanitizedExpandedContent, searchQuery)
-			: sanitizedExpandedContent
+		event.event_type !== 'tool_call' && event.event_type !== 'todo_update'
+			? searchQuery
+				? highlightHtmlContent(sanitizedExpandedContent, searchQuery)
+				: sanitizedExpandedContent
+			: ''
 	);
 
 	// Get event configuration
@@ -170,22 +174,27 @@
 		event.actor_type === 'subagent' && (!currentAgentId || event.actor !== currentAgentId)
 	);
 
-	// True when the event matched the search, but the match is buried in content
-	// that isn't shown on the collapsed card (e.g. full_content/full_thinking) —
-	// title + visible summary don't contain it, so there's nothing here to highlight.
+	// True when the event matched the search but the match is buried in content not shown on the collapsed card.
 	const hasHiddenMatch = $derived.by(() => {
-		// Once expanded, any match in full_content/full_thinking/etc. is already
-		// visible (rendered inline via ToolCallDetail or the generic markdown
-		// block below) — the dot would otherwise keep claiming it's still hidden.
+		// Expanded content is already visible via ToolCallDetail/the markdown block below.
 		if (isExpanded) return false;
 		const trimmedQuery = searchQuery.trim();
 		if (!trimmedQuery) return false;
-		const visibleSummary = event.summary
-			? isExpanded
-				? event.summary
-				: truncate(event.summary, 100)
-			: '';
-		const visibleText = `${displayTitle} ${visibleSummary}`.toLowerCase();
+
+		const visibleParts = [displayTitle];
+		if (event.summary && event.event_type !== 'todo_update') {
+			visibleParts.push(truncate(event.summary, 100));
+		}
+		if (toolName && event.event_type === 'tool_call') visibleParts.push(toolName);
+		if (shouldShowActorBadge) visibleParts.push(event.actor);
+		if (event.event_type === 'todo_update' && todosArray.length <= 3) {
+			// Mirrors TodoUpdateDetail's own inlineLimit (beyond 3 it shows a "click to see all" hint, not content).
+			for (const todo of todosArray) {
+				visibleParts.push(todo.content);
+				if (todo.status === 'in_progress' && todo.activeForm) visibleParts.push(todo.activeForm);
+			}
+		}
+		const visibleText = visibleParts.join(' ').toLowerCase();
 		if (visibleText.includes(trimmedQuery.toLowerCase())) return false;
 		return matchesSearch(event, searchQuery);
 	});
--- a/frontend/src/lib/components/timeline/TimelineEventCard.svelte
+++ b/frontend/src/lib/components/timeline/TimelineEventCard.svelte
@@ -17,6 +17,8 @@
 	import TodoUpdateDetail from './TodoUpdateDetail.svelte';
 	import ImageAttachments from '$lib/components/ImageAttachments.svelte';
 	import { markdownCopyButtons } from '$lib/actions/markdownCopyButtons';
+	import { highlightPlainText, highlightHtmlContent } from '$lib/utils/highlight';
+	import { matchesSearch } from '$lib/utils/timelineLogic.svelte';
 
 	interface Props {
 		event: TimelineEvent;
@@ -60,8 +62,10 @@
 
 	let isCopied = $state(false);
 
-	// Rendered markdown content for expanded view
-	let renderedExpandedContent = $state('');
+	// Sanitized markdown for expanded view — parsing/sanitizing is the expensive
+	// part, so this effect depends only on content/expansion state, NOT searchQuery.
+	// Highlighting is applied separately below as a cheap $derived over this.
+	let sanitizedExpandedContent = $state('');
 
 	// Render markdown when content changes or expansion state changes
 	$effect(() => {
@@ -71,19 +75,28 @@
 				event.metadata?.full_thinking ||
 				event.metadata?.full_text ||
 				event.metadata?.result_content ||
+				(event.metadata?.prompt as string | undefined) ||
 				event.summary ||
 				'';
 
 			const parsed = marked.parse(rawContent);
 			if (parsed instanceof Promise) {
 				parsed.then((html) => {
-					renderedExpandedContent = DOMPurify.sanitize(html);
+					sanitizedExpandedContent = DOMPurify.sanitize(html);
 				});
 			} else {
-				renderedExpandedContent = DOMPurify.sanitize(parsed);
+				sanitizedExpandedContent = DOMPurify.sanitize(parsed);
 			}
 		}
 	});
+
+	// Re-runs only the (cheap) text-node walk when searchQuery changes, not the
+	// full marked.parse + DOMPurify.sanitize pipeline.
+	const renderedExpandedContent = $derived(
+		searchQuery
+			? highlightHtmlContent(sanitizedExpandedContent, searchQuery)
+			: sanitizedExpandedContent
+	);
 
 	// Get event configuration
 	const isPlanEvent = $derived(
@@ -138,6 +151,7 @@
 			event.metadata?.full_thinking ||
 			event.metadata?.full_text ||
 			event.metadata?.result_content ||
+			event.metadata?.prompt ||
 			(event.summary && event.summary.length > 100)
 	);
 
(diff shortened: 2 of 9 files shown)


api/utils.py
```diff
@@ -749,6 +749,18 @@ def find_best_root(path: str, working_dirs: list[str]) -> str | None:
         return None
 
 
+# Cap on free-text tool-input fields (diffs, subagent prompts) placed into
+# timeline metadata. Without this, large Edit diffs or Task prompts ship in
+# full over the timeline endpoint, which live sessions poll every second.
+_METADATA_TEXT_LIMIT = 4000
+
+
+def _cap_text(value: str, limit: int = _METADATA_TEXT_LIMIT) -> str:
```
JayantDevkar, line 756: **Ask: add tests.** The suite (1796 tests) doesn't touch `get_tool_summary`'s new fields or `_cap_text` — I verified the behavior manually (exactly-4000 passes untouched; 9000-char input caps with the `[truncated, 9000 chars total]` marker; Edit carries `path`/`old_string`/`new_string`; Task carries `prompt`), but that evidence lives in a review, not the repo. A small parametrized test covering the boundary case and one capped case would lock in the contract the frontend now depends on.

Minor, while you're here: the marker text itself becomes searchable — a timeline search for "truncated" will match every capped event. Probably acceptable; flagging so it's a decision rather than an accident.

api/utils.py
```diff
@@ -780,7 +792,13 @@ def to_relative(path: str) -> str:
         return "Write file", to_relative(path), {"path": path, "content": content}
```
JayantDevkar, line 790: **Consider capping `content` here too.** `Write.content` still ships uncapped (pre-existing, not introduced by this PR), but it's now the single biggest remaining contributor to the 1s live-poll payload this PR's cap exists to bound — a session that writes a few large files carries all of them in full on every poll. `"content": _cap_text(content)` would close it; the display side already truncates at 2000 chars so nothing visible changes, and search coverage beyond 4000 chars is the same trade-off you already accepted for Edit diffs.

frontend/src/lib/utils/timelineLogic.svelte.ts
```diff
@@ -85,28 +85,92 @@ function matchesSingleFilter(
...
+ * hasHiddenMatch all used to run their own independent walk).
+ *
+ * Invalidated by a cheap content fingerprint, NOT object identity: a live
+ * session's poll (see ConversationView's POLL_INTERVAL_ACTIVE) replaces the
+ * entire event list with freshly-parsed objects every second even when
+ * nothing actually changed, so an identity check would rebuild every
+ * haystack every tick while a search is active. The fingerprint only needs
+ * to change when the *searchable content* changes, not the object reference.
+ */
+const haystackCache = new Map<string, { fingerprint: string; haystack: string }>();
```
JayantDevkar, line 119: **Ask: bound this cache.** Module-level `Map`, keyed by `event.id`, never evicted — and the haystack holds the event's *entire* lowercased metadata text (full thinking blocks, results, diffs), so a large session can pin several MB. Karma's main usage pattern is a dashboard tab left open all day navigating across sessions, so this accumulates every session ever searched for the lifetime of the tab.

Cheapest fix given SvelteKit reuses the module across client-side navigations: clear the cache when `createTimelineLogic` is instantiated (new session view → stale entries can never hit again, since ids are session-scoped), or an LRU/size cap if you want cross-navigation reuse. The fingerprint design itself is sound — this is only about lifetime.

frontend/src/lib/components/timeline/ToolCallDetail.svelte
```diff
@@ -240,13 +253,25 @@
 		return val != null && val !== '';
 	}
 
+	// Cap large free-text blocks (Write content, Edit diffs, Task prompts, the
+	// default JSON dump) before display/highlighting — these are only ever
+	// shown in a max-h-80 scroll box anyway, so highlighting the untruncated
+	// string (which can be multi-MB) wastes cycles clipped content never uses.
+	// Copy-to-clipboard buttons still use the original, untruncated string.
+	const DISPLAY_TRUNCATE_LENGTH = 2000;
```
JayantDevkar: **Truncation asymmetry with the server cap.** Server ships up to 4000 chars (`_METADATA_TEXT_LIMIT`) and search matches against all of it, but display clips at 2000 — so a query whose only match lives in chars 2000–4000 shows the amber "match in expanded content" dot, and then expanding shows no highlight anywhere. Same class of gap as the pre-existing 1000-char result display vs full-`result_content` search.

Either align this to 4000 (the strings are already server-bounded, so worst case is the same order of work), or keep 2000 and slice the displayed window around the first match. Tiny issue, but it lands exactly on the "why did this match?" confusion this PR set out to kill.

frontend/src/lib/components/timeline/TimelineEventCard.svelte
```diff
@@ -156,18 +170,26 @@
...
+		// visible (rendered inline via ToolCallDetail or the generic markdown
+		// block below) — the dot would otherwise keep claiming it's still hidden.
+		if (isExpanded) return false;
+		const trimmedQuery = searchQuery.trim();
+		if (!trimmedQuery) return false;
+		const visibleSummary = event.summary
+			? isExpanded
+				? event.summary
+				: truncate(event.summary, 100)
+			: '';
```
JayantDevkar: Nit: dead branch — this whole `$derived` early-returns `false` when `isExpanded` (line 180), so the inner `isExpanded ? event.summary :` ternary can never take the truthy arm. `const visibleSummary = event.summary ? truncate(event.summary, 100) : '';` says the same thing. No behavior change, just removes a head-scratcher.

frontend/src/lib/components/timeline/TimelineRail.svelte
```diff
@@ -204,11 +204,11 @@
...
 
-	// When search query changes, also update TimelineFilterBar's search if searchQuery is set externally
+	// Sync the external (Cmd+F) search query into the timeline's own search state,
+	// including when it's cleared — otherwise Escape closes the search bar but
+	// leaves every card's highlights and the filtered view stuck in place.
 	$effect(() => {
-		if (searchQuery) {
-			timeline.setSearchQuery(searchQuery);
-		}
+		timeline.setSearchQuery(searchQuery);
```
JayantDevkar, line 210: Informational, no change needed: the unconditional sync (including clears) is the right call, but note the external path it fixes is currently vestigial — `ConversationView` only ever *clears* `conversationSearchQuery` (Cmd+F toggles `showConversationSearch`, but no input is bound to the query), so this prop is always `''` today. The debounce-cancel effect in `TimelineFilterBar` guards the same not-yet-reachable path. Both are correct-by-construction for when the Cmd+F bar grows an input; flagging so nobody hunts for a search box that doesn't exist yet.

JayantDevkar (changes requested):
In-depth review: read the full diff, traced the data flow server→card→modal→tool panels, and verified the branch hands-on. The core work is solid and every claim in the PR description checked out — requesting changes only for a small set of scoped follow-ups below (tests + robustness), none of which are correctness bugs.

## What was verified

- **API**: `pytest` — 1796 passed; the 1 failure (`test_background_shells_router.py::test_empty_db_returns_empty_list`) is environmental — the suite reads the shared live `~/.claude_karma/metadata.db`, and the "extra" row it found was the reviewer's own test-runner shell. Also unit-exercised `_cap_text`/`get_tool_summary` directly: exactly-4000 string passes untouched, 9000-char `old_string` caps with the `[truncated, 9000 chars total]` marker, `path`/`prompt` fields intact.
- **Frontend**: `npm run check` — 0 errors, 13 warnings, all pre-existing and in files this PR doesn't touch.
- **In-browser** (session `6245d437`, 116 events): search filters + match counter work; the amber hidden-match dot appears exactly on cards whose match is in non-visible content and clears on expand; expanding shows the query highlighted inside the Edit ORIGINAL/REPLACEMENT panels; the previously blank "Editing:" path now renders (the `file_path`/`path` fix).
- **Escape-ordering fix confirmed empirically**: query `& health` (text contains a literal `&`) → 1 match, highlighted, no `&amp;` corruption — this exact query matched 0 events under the old escape-then-match code. Query `(archive` → 4 matches, no regex error (metacharacter escaping works).
- **Live-session cache fix confirmed**: searched a live session while it appended events on the 1s poll — query, highlights, and dot stayed stable as the total climbed 114→116. (A "3 shown vs 1 match" divergence observed during this test is pre-existing tailing behavior — `tailCount = 3` always shows the last 3 events alongside matches — not a PR bug.)

## Asks (all small)

1. Tests for the API metadata change (inline comment on `api/utils.py`).
2. Eviction/bound for the module-level haystack cache (inline comment on `timelineLogic.svelte.ts`).
3. Align display truncation (2000) with the server search cap (4000) or note the gap (inline comment on `ToolCallDetail.svelte`).
4. Consider capping `Write.content` too — it's the biggest remaining uncapped poll-payload contributor and this PR's stated goal (inline comment on `api/utils.py`).

Two further inline notes are informational only (vestigial Cmd+F path, dead ternary) — no action required for approval.

A deeper multi-agent review pass is still running; any additional confirmed findings will arrive as follow-up comments. Happy to re-review as soon as updates land.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

frontend/src/lib/components/timeline/TimelineEventCard.svelte
```diff
@@ -156,18 +170,26 @@
...
+		const visibleText = `${displayTitle} ${visibleSummary}`.toLowerCase();
```
JayantDevkar: **Bug: the dot fires on matches that are plainly visible.** `visibleText` counts only `displayTitle` + truncated summary, but the collapsed card also renders the tool-name badge (line ~314), the actor badge (~303), and — for todo events — the todo contents themselves via `TodoUpdateDetail`, *highlighted*, since it now receives `searchQuery`. Meanwhile the haystack includes `tool_name` and `actor`. Net effect, verified by code read: search `bash` → every Bash card shows its "Bash" badge *and* an amber dot claiming the match is hidden; search a subagent's name → same on its actor badge; search todo text on a collapsed ≤3-todo card → the match is visibly highlighted inline while the dot says it's hidden.

Fix direction: include `toolName`, `event.actor`, and (for todo_update events) the todo contents in `visibleText` — or simplest, treat todo_update cards as never-hidden since `TodoUpdateDetail` renders highlights in both collapsed and expanded modes.

frontend/src/lib/utils/timelineLogic.svelte.ts
```diff
@@ -85,28 +85,92 @@ function matchesSingleFilter(
...
+
+function getEventFingerprint(event: TimelineEvent): string {
```
JayantDevkar, line 124: **Bug (edge case): fingerprint misses late-arriving searchable metadata → permanently stale haystack.** `spawned_agent_slug` is merged from the `subagent_info` lookup (`api/services/conversation_endpoints.py:361`) and can flip from `null` to a value on a later poll while *every* fingerprinted field stays identical (same timestamp, summary length, result fields). It's not in `SEARCH_SKIP_KEYS`, so it belongs to the haystack — but the stale cache entry keeps serving the pre-slug text, so searching the agent's slug never matches the spawn event for the lifetime of the tab.

Cheap fix that closes the whole class rather than the instance: include a couple of structural signals in the fingerprint — e.g. `Object.keys(m).length` and `(m.spawned_agent_slug ?? '')` — or fingerprint the haystack's own length after a rebuild-on-poll for events whose `has_result` recently flipped. Any of these keeps the 1s-poll cache win while making "metadata grew a field" invalidate correctly.

frontend/src/lib/utils/timelineLogic.svelte.ts
```diff
@@ -85,28 +85,92 @@ function matchesSingleFilter(
...
+
+function getHaystack(event: TimelineEvent): string {
+	const fingerprint = getEventFingerprint(event);
+	const cached = haystackCache.get(event.id);
+	if (cached && cached.fingerprint === fingerprint) {
+		return cached.haystack;
+	}
+	const parts: string[] = [event.title, event.summary ?? '', event.actor];
+	collectSearchableText(event.metadata, parts);
+	const haystack = parts.join(' ').toLowerCase();
```
JayantDevkar: Nit: joining all fields with a single space lets a multi-word query match across the boundary of two unrelated fields (e.g. `old_string` ending in `foo` + `new_string` starting with `bar` matches the query `foo bar`) — a phantom match the old per-field matcher couldn't produce, and one that shows the hidden-match dot yet highlights nothing when expanded, since highlighting is per-field. Joining with `'\n'` doesn't fix it either — any separator that can appear in a query has this property, but a `'\u0000'` separator (which no realistic query contains) does. One-character change.

frontend/src/lib/components/timeline/TimelineEventCard.svelte
```diff
@@ -60,8 +62,10 @@
...
 
-	// Rendered markdown content for expanded view
-	let renderedExpandedContent = $state('');
+	// Sanitized markdown for expanded view — parsing/sanitizing is the expensive
+	// part, so this effect depends only on content/expansion state, NOT searchQuery.
+	// Highlighting is applied separately below as a cheap $derived over this.
+	let sanitizedExpandedContent = $state('');
 
 	// Render markdown when content changes or expansion state changes
 	$effect(() => {
```
JayantDevkar, line 69: Efficiency: this effect (and the `highlightHtmlContent` derived over it) runs for expanded **tool_call** events too, but its output is only rendered in the `{:else if}` branch — tool_call expansion always shows `ToolCallDetail` instead. Pre-existing waste, but the new `metadata.prompt` fallback raises the stakes: every expanded Task card now `marked.parse` + `DOMPurify`-sanitizes up to 4KB of prompt nothing displays, and re-walks it per debounced keystroke. Guarding with `event.event_type !== 'tool_call'` in the effect condition (and the derived) skips all of it. Same pattern applies to the popup path in `TimelineRail.svelte`.

frontend/src/lib/components/timeline/ToolCallDetail.svelte
```diff
@@ -457,8 +489,7 @@
 			<div
 				class="font-mono text-sm bg-[var(--bg-muted)] rounded-[var(--radius-md)] p-3 flex items-center justify-between gap-2"
 			>
-				<pre class="text-[var(--event-tool)] overflow-x-auto flex-1">$ {input.command ||
-						''}</pre>
+				<pre class="text-[var(--event-tool)] overflow-x-auto flex-1">$ {#if searchQuery}{@html highlightPlainText(String(input.command || ''), searchQuery)}{:else}{input.command || ''}{/if}</pre>
```
JayantDevkar: Consistency nit: the Bash command block is the only free-text block that skips `truncateForDisplay` — a command embedding a large heredoc gets fully escaped and regex-highlighted on every debounced query change, in a box with no max-height, while the sibling Write/Edit/Task/JSON blocks cap at 2000 chars. Wrapping it like its siblings (`truncateForDisplay(String(input.command || ''))`, keeping the copy button on the full string) closes the exact wasted-cycles case this constant was added to prevent.

frontend/src/lib/utils/highlight.ts
```diff
@@ -0,0 +1,97 @@
...
+		if (lastIndex < value.length) {
+			fragment.appendChild(document.createTextNode(value.slice(lastIndex)));
+		}
+		textNode.parentNode?.replaceChild(fragment, textNode);
+	}
+
+	// Re-sanitize: innerHTML was already sanitized once, but reading it back out
+	// and returning it as a fresh HTML string is a parse/serialize round-trip —
+	// re-sanitizing here is the cheap way to stay safe against that class of bug.
+	return DOMPurify.sanitize(container.innerHTML);
```
JayantDevkar, line 95: Optional: this re-sanitize doubles DOMPurify cost on the per-keystroke hot path this PR was otherwise careful to optimize, and the input is only ever already-sanitized HTML plus `<mark>` nodes built via `createElement`/`textContent` (never `innerHTML`), so there's no reachable injection for it to catch. Returning `container.innerHTML` directly halves the work. That said, defense-in-depth in front of `{@html}` is a defensible trade — your call; if you keep it, consider noting it's belt-and-suspenders so a future optimizer doesn't puzzle over it.

JayantDevkar (commented):
Follow-up to my earlier review — the deeper multi-agent pass finished and surfaced additional findings; each below was independently re-verified against the branch before posting. Two are real behavior bugs in the new feature (the hidden-match dot false-positive and the fingerprint staleness window); the rest are efficiency/consistency nits. The earlier CHANGES_REQUESTED verdict stands as the gate — treat the two bugs as part of that bar, the nits as judgment calls.

One finding has no diff line to anchor on: **`TimelineRail.svelte` ends with a stray markdown code fence** — the last line of the file is a literal ``` ``` ``` after the final `{/if}`, which Svelte renders as a text node at the bottom of every timeline. It's pre-existing on `main` (not caused by this PR), but this PR touches the file heavily, so sweeping it here is one deleted line.

Also a general style note rather than an inline thread: the multi-paragraph rationale comments added across this PR (the 14-line haystack-cache block in `timelineLogic.svelte.ts:124-137` is the largest; similar blocks in `TimelineEventCard`, `TimelineFilterBar`, `TimelineRail`, `ToolCallDetail`, `api/utils.py`) are written for the *reviewer of this diff* — cache-invalidation war stories, debounce race narratives. This repo keeps comments to short single-line pointers; rationale that justifies the change belongs in the PR description (where it already is, well written). Trimming each block to one line would match the surrounding code.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

api/utils.py
the-non-expert: Added `api/tests/test_utils_tool_summary.py` (10 tests) covering the boundary case (exactly-4000 passes untouched), the capped case (9000 chars → truncation marker), and that Edit/Write/Task carry their capped fields correctly. Left the marker-text-is-searchable behavior as-is per your framing — a decision, not an accident. Fixed in 474f36a.

api/utils.py
the-non-expert: Done — `content = _cap_text(tool_input.get("content", ""))` in the Write branch. Fixed in 474f36a.

frontend/src/lib/utils/timelineLogic.svelte.ts
the-non-expert: Went with the cheap option — `haystackCache.clear()` at the top of `createTimelineLogic`, since it's only ever instantiated once per session view and ids are session-scoped so stale entries could never hit anyway. Fixed in 474f36a.

frontend/src/lib/components/timeline/ToolCallDetail.svelte
the-non-expert: Aligned `DISPLAY_TRUNCATE_LENGTH` to 4000 to match the server's `_METADATA_TEXT_LIMIT`. Fixed in 474f36a.

frontend/src/lib/components/timeline/TimelineEventCard.svelte
the-non-expert: Removed — folded into the `hasHiddenMatch` rewrite for the false-positive bug below, which no longer has the dead ternary. Fixed in 474f36a.

frontend/src/lib/components/timeline/TimelineRail.svelte
the-non-expert: Acknowledged, no code change needed — resolving since you flagged this as informational only.

frontend/src/lib/components/timeline/TimelineEventCard.svelte
the-non-expert: Fixed — `hasHiddenMatch` now builds `visibleText` from title, summary (skipped for todo_update, which doesn't render it), toolName (when the badge shows), actor (when the badge shows), and todo contents when ≤3 todos render inline (mirroring `TodoUpdateDetail`'s own `inlineLimit`). Fixed in 474f36a.

frontend/src/lib/utils/timelineLogic.svelte.ts
the-non-expert: Went with your first suggestion — added `Object.keys(m).length` and `m.spawned_agent_slug ?? ''` to the fingerprint. Fixed in 474f36a.

frontend/src/lib/utils/timelineLogic.svelte.ts
the-non-expert: Switched the join separator to `String.fromCharCode(0)` (NUL) — one-character change as you noted. Fixed in 474f36a.

frontend/src/lib/components/timeline/TimelineEventCard.svelte
the-non-expert: Guarded both the effect and the derived with `event.event_type !== 'tool_call' && event.event_type !== 'todo_update'` in both `TimelineEventCard.svelte` and the popup path in `TimelineRail.svelte`. Fixed in 474f36a.

frontend/src/lib/components/timeline/ToolCallDetail.svelte
the-non-expert: Wrapped it — `truncateForDisplay(String(input.command || ''))`, copy button still uses the untruncated `input.command`. Fixed in 474f36a.

frontend/src/lib/utils/highlight.ts
the-non-expert: Kept the re-sanitize — didn't want to trade away defense-in-depth on an `{@html}` path for a minor perf win. Trimmed the comment to make clear it's an intentional belt-and-suspenders call, not leftover work. Fixed in 474f36a.

JayantDevkar (approved):
Re-verified at `474f36a` — every ask from both review rounds is addressed, and each fix was confirmed against the actual code and a live run, not just the replies. Approving.

## Verification of the fixes

- **API tests**: `tests/test_utils_tool_summary.py` — 10/10 pass, covering exactly the contract that matters (4000-boundary untouched, over-limit gets the marker, Edit/Write/Task carry capped fields). Full suite: **1806 passed** (up 10), same single pre-existing environmental failure (`test_background_shells_router` reads the shared live DB and found this reviewer's own shell — unrelated to the branch).
- **`Write.content` capped** — confirmed in `get_tool_summary`.
- **Haystack cache**: `haystackCache.clear()` on `createTimelineLogic` instantiation — correct choice given session-scoped event ids; fingerprint now includes `Object.keys(m).length` + `spawned_agent_slug`, closing the late-arriving-metadata staleness class, not just the instance.
- **Hidden-match dot**: `visibleText` now mirrors what the collapsed card actually renders (tool badge, actor badge, ≤3 inline todos matching `TodoUpdateDetail`'s `inlineLimit`). Verified in-browser: searching `bash` over a 116-event session → 13 matches, zero false dots on Bash-badged cards (previously every one showed the dot).
- **Truncation aligned** to 4000 = `_METADATA_TEXT_LIMIT`; Bash commands now truncate like their sibling blocks with copy preserving the full string.
- **Cross-field phantom**: NUL separator — verified one-char change.
- **Markdown pipeline** guarded for `tool_call`/`todo_update` in both the card effect/derived and the rail popup path.
- **Stray fence** removed from `TimelineRail.svelte`; comment blocks trimmed throughout; the kept re-sanitize in `highlight.ts` is clearly labeled intentional — reasonable call.
- `npm run check`: **0 errors**, same 13 pre-existing warnings in untouched files.

Nice turnaround — the fixes consistently addressed the *class* of each issue rather than the instance.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

PR #99, opened 2026-08-25
Settings v2: 7 new settings, write-safety hardening, schema-drift monitoring

the-non-expert:
## Summary
- Adds 7 new Claude Code settings to the dashboard (spell check, editor mode, notification channel, commit attribution, default model, update channel, verbose output), rendered generically from a manifest with restart badges, visible defaults, and reset-to-default.
- Hardens `api/routers/settings.py`'s write path — this dashboard writes to the real `settings.json` Claude Code reads at every session start: atomic per-request-unique temp file + `os.replace()`, serialized under an `asyncio.Lock` so concurrent writes can't corrupt the file, symlink-safe, single rolling backup, deep merge with proper null-deletes-key semantics, and jsonschema validation against a vendored settings schema.
- Adds a weekly GitHub Actions workflow that watches for upstream Claude Code settings schema drift and opens an issue/PR when a karma-exposed setting changes shape.
- Adds a Settings Guide doc page.

This branch went through two rounds of adversarial review focused specifically on the write path (concurrency, malformed/deeply-nested payloads, symlinked config files, schema-validation edge cases) — all findings from both rounds were fixed and independently re-verified against reproductions, not just the added tests.

## Test plan
- [x] `pytest api/tests/api/test_settings.py -v` — 21/21 passing, covering concurrent writes, null-on-absent-parent, typo'd keys inside strict nested schema objects, symlinked settings.json, and deeply-nested payloads
- [x] Full `pytest tests/` suite — passes except 5 pre-existing failures in `test_background_shells_router.py`, unrelated to this branch
- [x] Manual review at localhost (API :8020, frontend :5180) — settings page, About > Settings Guide


JayantDevkar (approved):
Approving. The write path is a real improvement over what's on `main` (shallow `dict.update()` + non-atomic `open("w")`, which could clobber nested keys and truncate on crash). I found no data-loss bug in the new path: symlink handling, permission preservation, unique temp file, rolling backup, null-deletes-key and the lock all read correct and are covered by the tests.

Follow-ups, in the order I'd do them (in the body rather than inline so they don't block the merge):

1. **Server-side allowlist of writable keys** — `api/routers/settings.py:303` keeps `extra: "allow"` and the vendored schema root is `additionalProperties: true`, so the endpoint will write any top-level key that validates, including ones Claude Code executes (`hooks`, `env`, `apiKeyHelper`). That's pre-existing on `main`, not introduced here, and `test_settings.py:286` documents it as deliberate scope — but the dashboard only ever needs the keys in `schema/watched-keys.json`. Restricting PUT to those, plus the narrow `:5180`-only origin gate #101 introduces, would make "write-safety hardening" complete. I'd like this one soon.
2. **`fsync` before `os.replace`** — `settings.py:117-127`. The docstring says a crash can never leave the file truncated; that holds for a process crash but not power loss (rename journaled, data blocks not). `f.flush(); os.fsync(f.fileno())` inside the `with`.
3. **Reset doesn't cancel a pending debounced save** — `+page.svelte` `handleManifestReset` never `clearTimeout(manifestTimeouts[entry.key])`. Type in "Default Model", hit reset within 500ms → `{model: null}` saves, then the orphaned timer rewrites the value you just cleared.
4. **Pin `peter-evans/create-pull-request@v6` to a SHA** — `.github/workflows/schema-drift.yml:101` runs with `contents: write` + `pull-requests: write`. The rest of the workflow is clean: schedule/dispatch triggers only, no `${{ }}` in `run:` bodies, remote text reaches `gh` only via `--body-file`.
5. Smaller:
   - `check_schema_drift.py:323` computes the digest from the trimmed schema, so every watched change opens two issues.
   - `:301` `ensure_ascii=True` vs `jq -S` raw UTF-8 will go permanently red on the first non-ASCII char upstream.
   - `StatusLineUpdate.type` allows `"disabled"` but the schema requires `const "command"`.
   - `jsonschema` is in `requirements.txt` but not `pyproject.toml`, and it's a hard import from `main.py`.
   - A JSONC `settings.json` (the schema allows trailing commas) 500s on GET/PUT — safe, but a dead end in the UI.

Not checked: the running UI, `schema/settings-reference.md` prose, Windows behaviour.

PR #100, opened 2026-08-28
Two-way terminal <-> dashboard links + resume button

the-non-expert:
## What this does

Two related features, both about connecting a terminal to the Karma dashboard in both directions:

1. **Terminal → Dashboard**: every session gets a short link (`http://localhost:5180/s/<8-char-id>`) that opens straight to that session's timeline. A `karma` shell script is included so you can jump from any terminal to its Karma page with one command, and `scripts/example-karma-statusline.py` is an optional copy-from reference if you want the link always visible in your status line — documented in SETUP.md, not auto-installed. Claude Code's statusLine is a personal, global setting shared across every project, so Karma never writes to it for you.

2. **Dashboard → Terminal (resume)**: the "resume" button on a session card, and on the session detail page, now actually does something instead of just copying a command. If the session is still running, it focuses that terminal (same as the existing focus-terminal feature). If the session has ended, it opens a new terminal, cd's into the right folder, and runs `claude --resume` for you.

## The link looks different depending on your terminal

Not every terminal can turn a short label into a clickable link the same way:

- **iTerm2, WezTerm, Kitty, Ghostty**: shows a real button-like link, e.g. "See this in Karma" — the actual URL is hidden underneath it.
- **Apple's Terminal.app**: shows a label followed by the plain URL. A plain click does nothing. Apple has never documented which modifier opens a link here, and it can vary — on the machine this was tested on, holding **Option and clicking once** opened it; if that doesn't work, try Cmd+double-click or right-click the URL and choose "Open URL".

This isn't a downgrade I chose. Apple's Terminal.app has never implemented OSC 8, the escape-code standard that lets a terminal turn a short label into a link with a different URL hidden underneath. Terminal.app can only make a URL clickable if the URL itself is visible on screen — there's no setting or software update that changes this. Confirmed in Claude Code's own docs. Switching to iTerm2 or another modern terminal is the only way to get the short button look.

## One thing to know before using this

These links only work if Karma is actually running when you click them:

- Either Karma's dev servers are up (`api` on 8020, `frontend` on 5180), or
- The Karma desktop app's "start at login" is turned on.

If neither is true, clicking a short link (or the resume button) just shows a browser error or a blank page, because there's no server to answer. That's expected for a locally-hosted dashboard, not a bug in this PR.

## Testing

- New API tests for the short-link resolver and the resume-or-focus endpoint, plus unit tests for the new terminal-launch service. All passing (full suite green aside from a pre-existing, unrelated flaky test file).
- Manually tested end to end on iTerm2 by the author: short link opens the right session's timeline, resume launches a new terminal and reconnects the session.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -154,6 +154,8 @@ Project (entry point)
 | GET | `/projects` | List all projects |
 | GET | `/projects/{encoded_name}` | Project details with sessions |
 | GET | `/sessions/{uuid}` | Session details |
+| GET | `/sessions/resolve/{short_id}` | Resolve a UUID or UUID prefix to its project (powers `/s/` short links) |
+| POST | `/sessions/{uuid}/resume-in-terminal` | Focus the session's terminal if live, else open a terminal running `claude --resume` (localhost-only) |
 | GET | `/sessions/{uuid}/timeline` | Event timeline |
 | GET | `/sessions/{uuid}/tools` | Tool usage |
 | GET | `/sessions/{uuid}/file-activity` | File operations |
@@ -183,6 +185,7 @@ Project (entry point)
 | `/plugins` | Plugins browser |
 | `/tools` | Tools browser |
 | `/sessions` | Global sessions browser |
+| `/s/[short_id]` | Short session link → redirects to the session's project page (statusline / `scripts/karma`) |
 | `/archived` | Archived sessions |
 | `/about` | About page |
 | `/settings` | User settings |
--- a/SETUP.md
+++ b/SETUP.md
@@ -449,6 +449,35 @@ Once the tracker hook is installed, live sessions show a `>_` button (session pa
 
 Sessions started before this feature gain the button automatically after their next activity (the hook backfills terminal identity on any event). Details: [docs/feature/open-terminal/spec.md](./docs/feature/open-terminal/spec.md).
 
+### Karma Link in Your Status Line (optional)
+
+Every session also gets a short link back to its Karma page: `http://localhost:5180/s/<8-char-id>`, which opens straight to that session's timeline. The link itself (`karma <short-id>`, or pasting the URL) needs only Tier 1 — it just reads Claude Code's own session files. Auto-detecting *which* session you're currently in needs Tier 2's live tracking, since that's what records which terminal each session runs in:
+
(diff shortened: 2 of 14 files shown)


JayantDevkar (approved):
Approving. The thing I cared most about here — an HTTP endpoint that spawns a terminal — holds up.

**Checked and clean**
- Command injection: `_build_command` `shlex.quote`s cwd and uuid, then `_applescript_str` escapes for AppleScript. Round-tripped 7 hostile paths (`"; do shell script "…"; "`, `$()`, backticks, embedded newlines) through real `osascript` — all came back byte-identical, payload stays inside single quotes.
- The request body carries nothing: uuid is gated by `_UUID_RE`, cwd comes only from on-disk session data, missing dir → 409.
- Cross-origin POST → 403 via the same `_reject_cross_origin` guard as `focus-terminal`.
- Short-id resolver: regex allows no glob metacharacters; redirect has a hardcoded `/projects/` prefix so no open redirect; `/resolve/{short_id}` isn't shadowed by `/{uuid}`.
- Non-darwin returns `launched=False` with a readable message, no crash. 24/24 new tests pass locally.

**Follow-ups (none blocking, leaving them here rather than inline)**
1. `api/routers/sessions.py:1927` — the origin guard trusts all of `cors_origins`, which includes `:3000`, `:3001`, `:5173`. Same guard as `focus-terminal`, so not a regression, but the impact went from "raise a window" to "start a process". #101 adds a narrower `rc_trusted_origins` (`:5180` only) for exactly this reason — once that lands, this endpoint should use the same gate.
2. `scripts/karma:40` — under `set -euo pipefail`, a failing `short_id="$(python3 …)"` kills the script before the dashboard fallback at `:66` (no `python3` on PATH → raw "command not found"). Also `os.path.getmtime(path)` at `:54` sits outside the `try`, so a live-session file deleted between `glob` and `getmtime` aborts the whole command. `short_id="$(python3 … || true)"` plus moving `getmtime` into the `try` fixes both.
3. `sessions.py:1935` — dedupe keys on the original uuid's live record. If `claude --resume` allocates a new session id (Karma's own chain model suggests it can), the original never goes live again and each click spawns another window. Not verified — I didn't run a real resume. A short per-uuid server-side cooldown would close it cheaply.
4. `sessions.py:1885-1896` — an ambiguous 8-char prefix silently resolves to the newest match. Fine today; worth returning a 409/chooser if it ever bites.
5. `frontend/src/routes/s/[short_id]/+server.ts:32` — unguarded `res.json()` turns a non-JSON 200 into a bare 500 instead of your 502.

Not checked: Windows/Linux at runtime, the iTerm2 branch at runtime, real `claude --resume` semantics.

PR #101, opened 2026-08-30
feat(live-sessions): per-session Remote Control toggle

the-non-expert:
## Summary

Adds a **Remote Control toggle** to the live-session header so you can turn Claude Code Remote Control on/off for a session straight from the Karma dashboard — no need to switch to its terminal to reach it from your phone.

Claude Code exposes no local API for Remote Control; the only lever is the `/remote-control` slash command typed **inside** the session. Karma runs on the same machine as the tracked terminal, so it drives Remote Control by typing into the session's terminal tab (tmux / macOS Terminal.app / iTerm2), reusing the tty / iTerm-session targeting from `services.terminal_focus`.

`/remote-control` is **not** a symmetric toggle, so the two directions differ:

- **ON** is fully automated — the bare command turns it on; the endpoint types it and confirms from the transcript.
- **OFF** opens Claude Code's interactive *"Disconnect Remote Control"* menu. Nothing drives that menu reliably from outside the process (`do script` can't deliver arrow keys; System Events needs a machine-wide Accessibility grant that misfires). So Karma types the command to *open* the menu, raises the terminal, and returns `method:"menu-open"` — the user picks *"Disconnect this session"* themselves, and the next `liveStatus` poll flips the pill.

State is read back from the session JSONL transcript, where Claude Code writes `{"type":"system","subtype":"bridge_status", …}` lines on connect/disconnect. No line anywhere → `off`; a line that can't be classified → `unknown` (endpoint 409s rather than blind-toggle).

The endpoint is **honest by construction**: `sent=false` means nothing was typed; `sent=true, confirmed=false` means keys went but the transcript hasn't caught up yet.

## Changes

- **`POST /live-sessions/{id}/remote-control {desired}`** — async, process-wide `asyncio.Lock` so button-mashing can't pile up keystroke sequences, bounded confirm-poll (6 × 0.5s).
- **`services/remote_control.py`** (new) — `read_remote_control_state`, `type_remote_control_command`, `can_send_remote_control`. Whole-transcript scan (48 MB cap) for the last `bridge_status`, **memoized on `(path, st_mtime_ns, st_size)`**; strict `type==system && subtype==bridge_status` match so ordinary text containing the string is ignored.
- **`RemoteControlToggle.svelte`** (new) — the pill: `off` / `on` (+ `↗ open` to claude.ai/code) / `unknown` / `busy` / `staged` / `menu` / `retry` / `ok`. Never dims; only the button disables. `role="switch"`, live-region announcements, reduced-motion.
- **`remote_control` + `can_remote_control`** on `LiveSessionSummary` — computed **only** on the single-session `GET` (they open the JSONL — too costly for the 1s-polled list endpoints).
- **CSRF gate** — `rc_trusted_origins` (default `:5180` only, deliberately narrower than `cors_origins` which also trusts `:5173`/`:3000`) + an `X-Karma-RC` request header, and **`Origin` must be present**. This required adding `X-Karma-RC` to `cors_allow_headers` so the browser's CORS preflight passes.
- **Status gate** — blocked only for `WAITING_INPUT` (a dialog is open; the trailing Enter could answer it), `STARTING` (no REPL yet) and `ENDED`. Everything else, including `ACTIVE`, is allowed — a slash command typed during a tool run just queues.
- **Hardening from a blind-review pass** (visible as `review #N` comments): `~/.claude/projects` path confinement against symlink escape, `https://claude.ai/` filter on the URL handed to `<a href>`, `session_id` regex, live-pid + supported-terminal gate.
- **`.gitignore`** — ignores `.storygraph.db` and its `-wal`/`-shm` sidecars. An external tool drops them in the repo root, nothing here references them, and they were showing up as untracked noise in every `git status`. (Unrelated to Remote Control, folded in as a drive-by.)
- `api-types.ts` carries two incidental Prettier line-collapses (`PermissionMode`, `CronStateTriggerEvent`) alongside the new RC types; `prettier --check` passes with them.

## Review round 2 — all five inline comments addressed

Rebased onto `main` (now includes #99, #100, #102); merges clean.

| Review comment | Resolution |
|---|---|
| `ConversationHeader.svelte` — conflicts with #100, `copiedTarget === 'resume'` becomes a type error | Kept **#100's `ResumeSessionButton` outright** and dropped this branch's copy-resume block. Taking the reviewer's note: #100's button focuses the terminal when live, which *is* useful, so this branch's "hide on live" was the wrong instinct and is gone. `CopyTarget` stays narrowed to `'uuid'`. |
| `live_sessions.py` — full-transcript scan on the 1s-polled single-session GET | Scan result memoized on `(path, st_mtime_ns, st_size)`. A JSONL transcript is append-only, so unchanged mtime+size ⇒ unchanged content and a quiet session is scanned exactly once. LRU-bounded at 256 entries, lock-guarded. **Measured — see Testing.** |
| `live_sessions.py` — a request with no `Origin` passes the gate | Missing `Origin` now 403s. For a keystroke-injecting endpoint it isn't worth the benefit of the doubt: this closes DNS-rebinding and the no-Origin-on-same-origin-POST corner. The dashboard always sends one; scripted callers can pass `Origin: http://localhost:5180`. |
| `remote_control.py` — "Esc-dismisses" overclaims; a stale-`off` ON click leaves the Disconnect menu open in an unfocused terminal | Took **both** options offered. The docstring no longer claims an Esc that nothing sends, and unconfirmed-ON now raises the terminal exactly like the OFF path, with a `detail` that says the disconnect menu may be showing. The frontend's `menu-open` method stays reserved for the OFF click, so the ON path reports the honest "sent, unconfirmed". |
| `remote_control.py` — tmux gate passes on pid liveness alone; recycled pid + recycled pane could enable RC on the wrong session | Implemented the suggested check: `tmux display-message -p -t <pane> '#{pane_pid}'` must be the recorded pid or one of its ancestors (walk capped at 8, 5s TTL cache). Unverifiable — no tmux, pane gone — is treated as **not** owned. **Verified against real tmux 3.7c** — see below. |

The reviewer's follow-up (moving #100's `resume-in-terminal` to `rc_trusted_origins`) is deliberately **not** in this PR — it hardens a different endpoint from a different feature, and was flagged as "follow-up once this lands". It'll be its own PR.

## tmux gate — verified on real tmux ✅

The pane-ownership check was initially mocked-only, which was called out as the one unproven claim in this PR. It is now verified end to end against **tmux 3.7c**, and the verification is committed as an integration test (`test_pane_ownership_against_real_tmux`, `skipif` when tmux is absent so CI and non-tmux machines are unaffected):

| Case | Expected | Result |
|---|---|---|
| `_tmux_pane_pid()` agrees with `tmux display-message` | equal | ✅ |
| The pane's own process | owned | ✅ |
| A **descendant** of the pane process — the real shape, since `claude` runs under the pane's shell | owned | ✅ |
| init (pid 1) | not owned | ✅ |
| An unrelated process outside tmux | not owned | ✅ |
| A nonexistent pane (`%99999`) | not owned | ✅ |

Run it yourself with `pytest tests/api/test_live_sessions_remote_control.py::test_pane_ownership_against_real_tmux -v` — it creates and destroys its own throwaway session and never touches your own. Locally it reports PASSED, not SKIPPED.

## Component

- [x] API (FastAPI backend)
- [x] Frontend (SvelteKit dashboard)
- [ ] Captain Hook (hook models)
- [ ] Hook Scripts
- [ ] Documentation
- [ ] CI / Build

## Type

- [x] Feature (new functionality)

## Testing

- [x] API tests pass — `test_live_sessions_remote_control.py` **40 pass** (33 + 7 new: pane-ownership accept/reject/unverifiable, a real-tmux integration test, scan memoization, per-path isolation, unconfirmed-ON raises vs confirmed-ON doesn't). Full `tests/api/` 1885 pass (the 6 `test_background_shells_router` failures are pre-existing, unrelated, and local-only: that test's fixture doesn't stub the router's `_maybe_sync()`, so each test ingests the developer's real `~/.claude/projects` history into its temp DB and gets 200 rows where it expects a handful. Green on CI, which has no such history. An earlier revision of this description blamed a stale `.storygraph.db` — that was wrong; karma's DB is `~/.claude_karma/metadata.db`.).
- [x] Frontend type-checks (`npm run check`, 0 errors) and lints clean (`npm run lint`, 0 errors)
- [ ] Captain Hook tests pass — N/A (no hook changes)
- [x] **Memoization measured, not asserted.** A/B across two detached worktrees (pre-fix vs post-fix), each a real uvicorn server hit over real HTTP with an isolated `$HOME`, against a 40 MB transcript containing **zero** `bridge_status` lines (the worst case: every byte scanned, nothing found). 100 sequential requests:

  | | pre-fix | post-fix | |
  |---|---|---|---|
  | Server CPU, read from `ps` (the OS, not the app) | 4.06 s | 0.06 s | **68× less** |
  | Wall clock, measured by the external request loop | 4.71 s | 0.63 s | 7.5× faster |

  That's **40.6 ms → 0.6 ms of CPU per request**. Both variants returned `remote_control: {'state': 'off'}`, confirming the scan path actually ran rather than an early-out. The regression test was mutation-checked: disabling the memo makes it fail `assert 3 == 1` for the right reason, so it detects the real defect instead of passing vacuously. (The endpoint's `@cacheable(max_age=5)` throttles how *often* a browser triggers this; it doesn't change the per-scan cost, which is what was measured.)
- [x] Manually tested — ON and OFF both verified end-to-end against a throwaway Terminal.app `claude` session:
  - **ON** → `{sent:true, confirmed:true, state:"on", url:"https://claude.ai/code/session_…"}`; confirmed three ways (endpoint response, a `bridge_status` "is active" line Claude Code wrote to the throwaway's own transcript, and a fresh Karma `GET` re-reading it as `on`).
  - **OFF** → `{sent:true, method:"menu-open", state:"on"}`; `focus_terminal` raised the window for the user to finish in the menu.

## Screenshots

_Pill states (off / on+open / menu / retry) render via `npm run check` + component review; no static screenshots attached._

## Related Issues

Separate feature branch off `main`; independent of #99 (settings / schema-drift). Rebased onto `main` after #99/#100/#102 merged.


frontend/src/lib/components/conversation/ConversationHeader.svelte
```diff
@@ -494,22 +505,25 @@
...
-							<Terminal size={11} strokeWidth={2} />
-							{#if copiedTarget === 'resume'}
-								<span>copied!</span>
-							{:else}
-								<span>resume</span>
-							{/if}
-						</button>
+						<!-- Copy resume command — only useful once the session has ended;
+						     while it's live you'd just go to its terminal. -->
+						{#if !liveStatus || liveStatus.status === 'ended'}
```
JayantDevkar, line 510: Conflicts with #100 (merged). #100 replaced this copy-button block with `<ResumeSessionButton>` behind `showResume` and narrowed `CopyTarget` to `'uuid'`, so `copiedTarget === 'resume'` becomes a type error if the conflict is resolved by keeping both. On rebase, keep #100's button and apply your live-session condition to it — though note #100's button focuses the terminal for live sessions, so hiding it on live may no longer be what you want.

api/routers/live_sessions.py
```diff
@@ -513,7 +542,7 @@ def get_live_session(
             detail=f"Live session not found: {session_id}",
         )
 
-    return state_to_summary(state)
+    return state_to_summary(state, include_remote_control=True, projects_dir=config.projects_dir)
```
JayantDevkar, line 545: This endpoint is polled by `ConversationView.svelte` at `POLL_INTERVAL_ACTIVE = 1000`ms. With `include_remote_control=True`, each uncached hit runs `_last_bridge_status_in`, which reads the whole transcript (up to 48 MB) into memory — and in the common case (RC never used, so no `bridge_status` line anywhere) it scans every file in the chain. `@cacheable(max_age=5)` softens it to ~1 scan / 5s per open tab, but on a long session that's still a sustained multi-MB read + decode plus a large allocation each time. Suggest memoizing the result on `(path, st_mtime, st_size)` so an unchanged transcript is never re-read.

api/routers/live_sessions.py
```diff
@@ -571,6 +600,245 @@ def focus_session_terminal(
...
+    """Stricter CSRF gate for the keystroke-injecting toggle (review #4).
+
+    ``cors_origins`` also trusts :5173/:3000 — any local dev server. Enabling
+    Remote Control on the user's live sessions must be reachable only from
+    Karma's own dashboard, so this checks ``rc_trusted_origins`` and also
+    requires a custom header (which a cross-origin simple request cannot set,
+    and setting it forces a CORS preflight the middleware then screens).
+    """
+    origin = request.headers.get("origin")
+    if origin is not None and origin not in config.rc_trusted_origins:
```
JayantDevkar, line 645: A request with no `Origin` header passes this check. The `X-Karma-RC` requirement still stops cross-origin browser requests, so this is low risk — but for a keystroke-injecting endpoint I'd require `Origin` to be present and trusted, which also closes the DNS-rebinding + no-Origin-on-same-origin-POST corner. The dashboard always sends it.

api/services/remote_control.py
```diff
@@ -0,0 +1,365 @@
...
+) -> Dict[str, Any]:
+    """Current Remote Control state for a session.
+
+    Returns ``{state: "on"|"off"|"unknown", url: str|None, at: str|None}``.
+
+    - **No** ``bridge_status`` line anywhere in the chain → ``"off"``. Remote
+      Control has to be turned on explicitly and writes a line when it does, so
+      "never seen one" means off. (The rare case where it was on via
+      ``--remote-control`` and the line has scrolled past the tail degrades
+      safely: turning "on" again just opens — and Esc-dismisses — the menu; we
```
JayantDevkar, line 187: Nothing in this module sends Esc, so "Esc-dismisses" overclaims. On a stale-`off` reading, `desired == "on"` leaves the Disconnect menu open in an unfocused terminal, and the user's next Enter there picks a menu item. Either fix the docstring or raise the terminal in that case like the OFF path does.

api/services/remote_control.py
```diff
@@ -0,0 +1,365 @@
...
+)
+_ON_MARKERS = ("is active", "session is active", "remote control is on")
+
+
+# ---------------------------------------------------------------------------
+# Capability gate
+# ---------------------------------------------------------------------------
+
+
+def can_send_remote_control(terminal: Optional[Dict[str, Any]]) -> bool:
```
JayantDevkar, line 86: For tmux this returns True on pid liveness alone, and `pid_is_live_claude` accepts `node`/`bun`/`deno`. A recycled pid + recycled pane id would pass, and if that pane is now a *different* Claude session, this enables RC on the wrong session. Cheap check: confirm `tmux display-message -p -t <pane> '#{pane_pid}'` is the recorded pid or its ancestor.

JayantDevkar (changes requested):
The security design here is solid and I'd like to land this — but it needs a rebase and one perf fix first.

**Checked and clean**
- The typed payload is the constant `RC_COMMAND`; nothing request- or session-derived reaches it. All subprocess calls are list-form, AppleScript interpolation goes through `_applescript_str`, tmux uses `send-keys -l --`.
- The origin gate is exact-match and enforced server-side independent of CORS, so adding `X-Karma-RC` to `cors_allow_headers` doesn't let `:5173`/`:3000` through (they pass preflight, then get 403). Desktop app uses `:5180`, so it's covered.
- Transcript paths are realpath-confined under `projects_dir`; the `href` is filtered to `https://claude.ai/` on both sides. Blocking calls in the async handler are all in `asyncio.to_thread`.

**Needs to change before merge**
1. Rebase: #100 just merged and this now conflicts in `ConversationHeader.svelte` and `api/schemas.py` (see inline).
2. The full-transcript scan on the polled single-session GET (see inline).

Follow-up once this lands: #100's `resume-in-terminal` endpoint should move to this PR's narrower `rc_trusted_origins` gate — it spawns a process and currently trusts all of `cors_origins`.

Not checked: real toggle against a live terminal, Claude Code's actual `bridge_status` wording vs `_ON_MARKERS`/`_OFF_MARKERS` (a wording change degrades to `unknown` → 409, which is the safe direction).

frontend/src/lib/components/conversation/ConversationHeader.svelte
the-non-expert: Rebased onto `main`. Kept **#100's `ResumeSessionButton` outright** and dropped this branch's copy-resume block entirely — your note landed: #100's button focuses the terminal when live, which *is* useful, so this branch's "hide on live" was the wrong instinct. `CopyTarget` stays narrowed to `'uuid'`, so no type error. Fixed in a08634c.

api/routers/live_sessions.py
the-non-expert: Memoized exactly as suggested — on `(path, st_mtime_ns, st_size)`, LRU-bounded at 256, lock-guarded. A JSONL transcript is append-only, so unchanged mtime+size means unchanged content and a quiet session is scanned once.

Measured rather than asserted: A/B across two detached worktrees (pre-fix vs post-fix), each a real uvicorn server over real HTTP with an isolated `$HOME`, against a 40 MB transcript with **zero** `bridge_status` lines — the worst case you described. 100 sequential requests:

| | pre-fix | post-fix | |
|---|---|---|---|
| server CPU, read from `ps` (OS, not the app) | 4.06 s | 0.06 s | **68× less** |
| wall clock, external request loop | 4.71 s | 0.63 s | 7.5× |

40.6 ms → 0.6 ms of CPU per request. Both variants returned `remote_control: {"state": "off"}`, confirming the scan path ran. The new regression test was mutation-checked — disabling the memo makes it fail `assert 3 == 1` for the right reason. Fixed in a08634c.

api/routers/live_sessions.py
the-non-expert: Agreed — missing `Origin` now 403s rather than passing. For a keystroke-injecting endpoint it isn't worth the benefit of the doubt, and it closes the DNS-rebinding + no-Origin-on-same-origin-POST corner you named. The dashboard always sends one; scripted callers can pass `Origin: http://localhost:5180` explicitly. Test `test_no_origin_allowed` inverted to `test_no_origin_403`. Fixed in a08634c.

api/services/remote_control.py
the-non-expert: Took **both** options rather than either. The docstring no longer claims an Esc that nothing here sends, and the stale-`off` case now behaves: unconfirmed ON raises the terminal exactly like the OFF path, with a `detail` saying the disconnect menu may be showing instead. The frontend's `menu-open` method stays reserved for the OFF click, so the ON path still reports the honest "sent, unconfirmed" rather than mislabelling itself. Covered by `test_enable_sent_but_unconfirmed_raises_terminal` and `test_enable_confirmed_does_not_raise_terminal` (the happy path must not steal focus). Fixed in a08634c.

api/services/remote_control.py
the-non-expert: Implemented exactly the check you suggested: `tmux display-message -p -t <pane> '#{pane_pid}'` must be the recorded pid or one of its ancestors (walk capped at 8 levels, 5s TTL cache since this runs on a polled endpoint). Unverifiable — no tmux, pane gone, `ps` unavailable — is treated as **not** owned, so the gate closes rather than typing blind.

⚠️ **Leaving this thread open deliberately.** tmux isn't installed on my dev machine, so that invocation has never actually been executed — it's covered only by mocked unit tests. The failure direction is safe (unverifiable ⇒ toggle withheld, never mis-typed), but if the command string is wrong in any way the toggle would silently vanish for **every tmux user** while looking perfectly fine on Terminal.app / iTerm2.

If you have tmux handy, confirming the pill still appears for a tmux-hosted session would close this. I'd rather flag it than mark it done. a08634c.

the-non-expert: Update: **verified against real tmux (3.7c)** — no longer taking my word for it. Installed tmux and drove the gate end to end:

```
_tmux_pane_pid('%0') -> 97813   (tmux says 97813)
  [PASS] pane's own pid is owned
  [PASS] descendant pid is owned
  [PASS] init (pid 1) is NOT owned
  [PASS] this unrelated python is NOT owned
  [PASS] nonexistent pane is NOT owned
```

The third line is the one that mattered: `pane_pid` is the pane's shell and `claude` runs underneath it, so the ancestor walk is what the real setup actually depends on.

Committed as an integration test in 3c429e9 rather than left as a one-off probe — `test_pane_ownership_against_real_tmux` creates and destroys its own throwaway session, and `skipif`s when tmux is absent so CI and non-tmux machines are unaffected. It reports PASSED (not SKIPPED) on a box with tmux:

```
pytest tests/api/test_live_sessions_remote_control.py::test_pane_ownership_against_real_tmux -v
```

Resolving this now that the invocation is confirmed rather than assumed.

PR #102, opened 2026-09-12
Fix hook timeout units: seconds, not milliseconds (#98)

### LINKED ISSUE #98: SETUP.md documents hook `timeout` as milliseconds, but it is seconds (5000 = 83 minutes)
## Summary

`SETUP.md` states that Claude Code hook `timeout` values are in **milliseconds**. They are in **seconds**. Every hook config the guide tells users to paste therefore installs a timeout ~1000x larger than intended.

## The incorrect statement

`SETUP.md:379`:

> **Important:** Timeout values are in **milliseconds** for command hooks (5000 = 5 seconds).

## What the official docs say

From the [Claude Code hooks reference](https://code.claude.com/docs/en/hooks), the `timeout` field is:

> **Seconds** before canceling. Defaults: 600 for `command`, `http`, and `mcp_tool`; 30 for `prompt`; 60 for `agent`.

So the recommended `"timeout": 5000` is not 5 seconds — it is **5000 seconds, or 83 minutes**. A hung hook that should be cancelled after 5s instead blocks for well over an hour.

## Scope

This is not limited to the one sentence. `SETUP.md` contains **21** affected values:

- 18 x `"timeout": 5000` (intended: 5)
- 3 x `"timeout": 15000` (intended: 15)

`captain-hook/README.md:101` also carries `timeout: 5000` in an example config. Note that the same file is already correct in prose at line 152 ("Default 60s for commands, 30s for prompt-based hooks"), so the repo currently contradicts itself.

## Impact

Anyone who followed `SETUP.md` has eight hook registrations with an 83-minute cancellation window. It is silent — nothing surfaces until a hook actually hangs, at which point the session appears frozen rather than recovering.

## Suggested fix

1. Correct `SETUP.md:379` to say **seconds**.
2. Change the 18 `5000` values to `5` and the 3 `15000` values to `15`.
3. Change `captain-hook/README.md:101` from `timeout: 5000` to `timeout: 5`.

Happy to open a PR if useful.

the-non-expert:
## Summary

Claude Code's hook `timeout` field is in **seconds**, not milliseconds. `SETUP.md` and captain-hook's docs told users to install `"timeout": 5000` (intended as 5s), which actually installs an **83-minute** cancellation window instead. This was live on real systems, not just theoretical — verified on two machines where karma-installed hooks carried the exact `5000`/`15000` values straight from the docs.

- Corrects all 144 example timeout values (and the prose claiming milliseconds) across `SETUP.md`, `api/SETUP.md`, `api/scripts/SETUP.md`, `captain-hook/README.md`, `captain-hook/docs/hooks/*`, and `docs/about/hooks-guide.md`.
- Fixes a related **real code bug** this same confusion caused in karma's own `/hooks` browser page: `HookRegistration.timeout_ms` stored the raw (seconds) value from `settings.json` with no conversion, then displayed it suffixed `ms` — silently reinforcing the same mistake for anyone using karma to audit their own hooks. Renamed to `timeout_seconds` end-to-end (Pydantic model, `api-types.ts`, hooks page template).

## Test plan
- [x] `pytest -k hook` passes (no dedicated `hook.py` test file exists; ran shell-cron tests to confirm the unrelated `timeout_ms` field on `background_shell` was untouched)
- [x] Sanity-instantiated `HookRegistration` with `timeout_seconds` and confirmed `routers/hooks.py` still imports cleanly
- [x] Grepped repo-wide to confirm no remaining `5000`/`15000`-style ms-mistake values or stray `timeout_ms` references tied to hooks

Fixes #98

https://claude.ai/code/session_01Gv24uWUqNLigGcTtnGEHQv


docs/feature/hooks-page/plan.md
```diff
@@ -156,7 +156,7 @@ HookRegistration
 │   ├── command: str                 # Full command string
 │   ├── script_path: str             # Resolved path to script file
 │   ├── script_language: enum        # "python" | "node" | "shell" | "unknown"
-│   ├── timeout_ms: int              # Timeout in milliseconds
+│   ├── timeout_ms: int              # NOTE: misnamed -- holds raw seconds from settings.json, never multiplied; rename to timeout_seconds
```
JayantDevkar, line 159: Nit: this PR already does the rename, so the note is stale on merge. Could just be `timeout_seconds: int  # Timeout in seconds (raw value from settings.json)`. Same for the "despite the field naming below" aside on line 82.

JayantDevkar (approved):
Verified against the Claude Code hooks reference: `timeout` is seconds, so the old docs really did install 83-minute windows. Approving.

What I checked:
- All 136 changed doc values are an exact ÷1000 (5000→5, 2000→2, 15000→15, …) — no stray edits.
- The Bash `tool_input.timeout: 120000` example is correctly left in ms (good catch in the second commit).
- `timeout_ms` → `timeout_seconds` is renamed end-to-end (model, parser, `api-types.ts`, hooks page). The remaining `timeout_ms` refs are `background_shells`, which is genuinely ms — untouched, correct.
- No installer/script in the repo writes hook timeouts, so docs were the only source of bad values.

Non-blocking follow-ups:
- `docs/feature/hooks-page/ui-design.md` (not in this diff) still shows `timeout: 5000ms` / `15000ms` / `30000ms` in the mockups at lines 223, 231, 240, 284.
- Users who already pasted the old config still have 5000s timeouts. A one-line note in SETUP.md (or a check on the /hooks page flagging timeouts > 600) would help them find it.

